PR #1, opened 2017-01-19
Initial outline for specification

jonasfj:
@hammad13060, @djmitche, @imbstack,
An initial outline for the `json-e` spec... We need a LOT more features, so write a section and file a PR :)
The file might get so long that we need a file per section... but for now this works.

I think we should do specification changes in a PR before implementing a feature, this way we can discuss features by example before we implement them. And we have test cases when we start implementing them.

@hammad13060, please accept invitation to join the organization (it requires you enable 2FA if not already).

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -1,0 +1,153 @@
+################################################################################
+# ### Specificaiton for json-e
+# This document specifies the json-e template language through a series of
+# test cases on the form: `{title, context, template, result}`
+# Where `jsone.render(template, context) = result` and title is short name
+# for the test case. If no result is specified, it's because the test cases must
+# throw an error.
+#
+# Comments explaining the test cases is encouraged, additionally entries with
+# {section: '...'} is encouraged to separate sections.
+#
+# Test conditions:
+#   1) Context is assumed to always functions:
+#     - max(a, b)
+#     - min(a, b)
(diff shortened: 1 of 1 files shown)


hammad13060:
@all 

I have updated [json-e](https://github.com/hammad13060/json-e) for following features

1. ${...} for string interpolation
2. $eval for object replacement
3. $if construct : $then and $else support evaluation of 1 and 2 mentioned above. If we have something like $then: <<<some js object>>> it will be recursively evaluated.
2 $switch construct: It supports 1 and 2 mentioned above along with recursive object evaluation like in if-else construct.

jonasfj:
@hammad13060, I propose that we proceed by:
 A) gathering consensus on syntax by changing `specification.yml` in this repo through PRs
 B) submitting PRs to this repo with implementation of features from the specification

That way the code goes through a review process.

jonasfj:
@hammad13060 anything controversial in this PR? if not I suggest we land it... So we make PRs ontop of it rather than copying eveyrthing into new proposals.

Ie. suggesting changes/additions to the spec is perfectly fine.

hammad13060:
I have gone through it and it seems good. Yes lets land it (that will be more efficient).

PR #2, opened 2017-01-19
spec 2 proposal with different $fromNow format

hammad13060:
@jonasfj @djmitche and @imbstack 

I propose following specs for ```$fromNow```

```javascript
{
  $fromNow: {
      d/day/days: <some days>,
      h/hour/hours: <some hours>,
      m/min/minutes: <some minutes>
   }
}
```

So we will have something like this

```javascript
{
  $fromNow: {d: 2, h: 12, m: 32},
  $fromNow: {day: 2, hour: 12, min: 32},
  $fromNow: {days: 2, hours: 12, minutes: 32},
}
```

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -1,0 +1,153 @@
+################################################################################
+# ### Specificaiton for json-e
+# This document specifies the json-e template language through a series of
+# test cases on the form: `{title, context, template, result}`
+# Where `jsone.render(template, context) = result` and title is short name
+# for the test case. If no result is specified, it's because the test cases must
+# throw an error.
+#
+# Comments explaining the test cases is encouraged, additionally entries with
+# {section: '...'} is encouraged to separate sections.
+#
+# Test conditions:
+#   1) Context is assumed to always functions:
+#     - max(a, b)
+#     - min(a, b)
+#   2) Time at test time is expected to be '2017-01-19T16:27:20.974Z'
+#   3) JSON is rendered to strings without whitespace and preserving order of
+#      keys
+################################################################################
+---
+section:  Initial Minimal Subset
+---
+title:    Identity
+context:  {}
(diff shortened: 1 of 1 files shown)


jonasfj:
I like this a lot... That we can also have $eval inside the parameters for
$fromNow.

The only argument for the string syntax is that it's shorter... We have the
parser for it implemented in taskcluster-client.

Hmm, perhaps we can have both... Parse string if given other use object...

Den 19. jan. 2017 1.42 PM skrev "Hammad Akhtar" <notifications@github.com>:

hammad13060:
ya keeping both seems good :)

jonasfj:
r+

But let's keep the string syntax too. It's easy to parse with a regexp.

hammad13060:
I was also thinking that instead of having ```d/day/days```, ```h/hour/hours``` and ```m/min/minutes``` lets limit ourselves to ```d, h, m``` only

jonasfj:
In the string format I really like having the ability to write `days` and `day` it so nice to read...

hammad13060:
for the string part lets keep all the variants, however for time specification using object, lets limit ourselves to to d, m, h.

jonasfj:
okay, I can support that...

jonasfj:
actually @imbstack had a good point that 99% of the time the `$fromNow` construct is used with a string like: `'1 day'` or `'20 minutes'`, so maybe this object support is overkill.

If people want to pass in variables, they can use string interpolation, but I'm guessing that 99% of time it'll be faster and shorter to just write `{$fromNow: '2 hours'}`

hammad13060:
Okay. Then probably we should go with the string format only.

jonasfj:
Yeah, let's stick with string for now...

PR #3, opened 2017-01-19
Proposed simple expressions

jonasfj:
I think we can make an outright crazy amount of test cases... We probably should make some really complex one that tests multiple features in one case :)

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -151,3 +151,36 @@ context:  {}
 template: {$fromNow: '1 day'}
 result:   '2017-01-20T16:27:20.974Z'
 ################################################################################
+---
+section:  expression language
+---
+title:    addition
+context:  {a: 1, b: 2}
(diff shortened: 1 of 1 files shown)


hammad13060:
This looks nice. However I am also thinking about cases where ```${....}``` will evaluate expressions (just like those you mentioned) and then it will perform interpolation.

jonasfj:
Absolutely, feel free to add more examples. This doesn't even cover expressions like `2 + 3 * 4` which we properly need to test with too :)

jonasfj:
@hammad13060, I've enabled branch protection, so you need to click the review button before we can land this...

IMO, please feel free to add similar examples.. Let's build up a huge spec/list-of-test-cases before we get too far with implementing it.... Then we can implement one feature at the time and see the number of failing testcases decline :)

hammad13060:
Yup, I will add more test cases to the specs.

PR #4, opened 2017-01-21
new specifications for $if and chaining of nested objects in context

hammad13060:


specification.yml
```diff
@@ -102,6 +102,56 @@ title:    $if -> delete-marker, false
 context:  {cond: false}
 template: {key: {$if: 'cond'}, k2: 3}
 result:   {k2: 3} # missing then/else branches should return a delete-marker
+---
+title: $if->then, then => $eval, true
+context: {key: {b: 1}}
+template: {$if: true, then: {$eval: 'key'}}
```
jonasfj: I think this should be `{$if: 'true', ...}`.... Just to ensure YAML doesn't parse 'true' as a boolean.

You want it to be a string right?

specification.yml
```diff
@@ -184,3 +234,21 @@ context:  {a: 3, b: 2}
...
+section: accessing nested objects => context
+---
+title: $eval
+context:  {key: {key2: ['value', 1, 2, true, {}]}}
+template: {$eval: 'key.key2'}
+result:   ['value', 1, 2, true, {}]
+---
+title:    $dumps
+context:  {key: {key2: [1,2,true,{},[]]}}
+template: {$dumps: 'key.key2'}
```
jonasfj: Disagree, $dumps doesn't evaluate an expression.... It evaluates JSON-e... So you need a level of {$eval} inside $dumps....

hammad13060: Can you explain what ```$dumps``` do in detail. As far as I understand it fetches an object from context and converts it into a string representation.

hammad13060:
I will file a separate PR for array access. We will cover both index access and array slicing in that.

I am thinking about [this](https://github.com/hammad13060/json-e#array-access) kind of array access, because this way we can add support for slicing very easily.

specification.yml
jonasfj: Look at the code samples... Already I the spec.
We have two kinds of expressions:
 1. JSON-e expressions
 2. Expression based language

(1) is things like {$dump: ...}, {$if: ...}, Any JSON value , {$eval: ... }, etc...

(2) is expressions like ' 1 + a' IE stuff that is parsed...

The argument for $dump is of type 1. Whereas the argument for {$eval: ...} is type 2.

jonasfj: Effectively, the $eval construct bridges between the two types...

specification.yml
```diff
@@ -41,8 +41,8 @@ template: {$eval: 'key'}
 result:   ['value', 1, 2, true, {}]
 ---
 title:    $dumps
-context:  {key: [1,2,true,{},[]]}
-template: {$dumps: 'key'}
+context:  {}
+template: {$dumps: [1,2,true,{},[]]}
```
jonasfj: Ah, nice fix.. I probably did it wrong...

PR #6, opened 2017-01-22
A few more $dumps examples

jonasfj:
Does this make the plan for `$eval` more obvious?


hammad13060:
Yes this is more informative 👍

PR #7, opened 2017-01-22
Proposal for $map

jonasfj:
I'm definitely open to other ideas...
Most the ones I had was something like:
```js
{
  $for: 'x' # specify variable
  in: [1,2,3]
  each: {$eval: 'x + 1'}
}
```
But it feels long an hard to see that `x` is the variable that gets declared in the context for the evaluation of the json-e expression given in `each`.

I like using `each(x)` because it's easy to read that each entry in the array will have the name `x` in the subexpression...

Thoughts, please? Other ideas are welcome too :)

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -184,3 +184,58 @@ context:  {a: 3, b: 2}
 template: {$eval: '"" + (a + b)'}
 result:   '5'
 ################################################################################
+---
+section:  $map
+---
+title:    simple map
+context:  {a: 1}
+template:
+  $map: [2, 4, 6]
+  each(x): {$eval: 'x + a'}
+result:   [3, 5, 7]
+---
+title:    complex identifier
+context:  {a: 1}
+template:
+  $map: [2, 4, 6]
+  each(my_identifier97): {$eval: 'my_identifier97 + a'}
+result:   [3, 5, 7]
+---
+title:    must wrap expressions with $eval
+context:  {a: 1}
+template:
+  $map: [2, 4, 6]
+  each(my_identifier97): 'my_identifier97 + a'
+result:   ['my_identifier97 + a', 'my_identifier97 + a', 'my_identifier97 + a']
+---
+title:    can take from objects
+context:  {a: 1}
+template:
+  $map: [{k: 1}, {k: 2}, {k: 3}]
+  each(y): {$eval: 'y.k'}
+result:   [1, 2, 3]
+---
+title:    can make objects too
+context:  {a: 1}
+template:
+  $map: [{k: 1}, {k: 2}, {k: 3}]
+  each(y):
+    k: {$eval: 'y.k + 1'}
+    v: 'before=${y.k}'
+result:
+  - k: 2
+    v: 'before=1'
+  - k: 3
+    v: 'before=2'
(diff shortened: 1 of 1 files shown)


hammad13060:
```each(x)``` syntax looks better. However, polluting the context with ```x``` may cause some issues like someone may overwrite some useful property in context which is also represented by key ```x```'. So I think before evaluating ```$map``` construct we should check if key already exist in the context, If so we must throw an error.

**Comment by hammad13060** [comment]
Probably you can add some examples which should throw errors.

**Comment by jonasfj** [comment]
Well, people can choose another variable name than `x` if they want... Otherwise it's no different than normal scope rules... IE. x will only be overwritten inside the sub-jsone-expression...

Similar to JS like this:
```js
let x = 2;
let f = () => {
  let x = 7;
  console.log(x); // writes 7
}
f();
console.log(x); // writes 2
```

hammad13060:
Yes agreed that people can choose another variable name, but what if that variable already exist as a property and by mistake user decided to use that variable name for ```$map``` construct.

jonasfj:
Yeah, normally a compiler could warn about such overwrites but that's
unlikely here.

Honestly it would be nice to protect against such overwrites. But in
practice if we did use this for tc-hooks it would mean that we couldn't add
new variables to the context in the future. As such a variable might
conflict with an existing each(..) identifier, and such a conflict would
break a template that otherwise works.

IE. If overwrites aren't allowed adding a new context variable breaks
compatibility for existing templates. That's bad, as existing templates
would work perfectly fine, they wouldn't use the new variable. But as long
as we don't remove variables from tc-hooks, we can always add new ones.
Unless overwrites are forbidden.

Den 22. jan. 2017 12.03 AM skrev "Hammad Akhtar" <notifications@github.com>:

hammad13060:
probably in specs we can specify that variables for ```$map``` should use ```$``` as prefix.

jonasfj:
In that case I would be tempted to hardcore the variable name as $$, but I
don't think overwrites are bad... It's pretty obvious.

Den 22. jan. 2017 1.06 AM skrev "Hammad Akhtar" <notifications@github.com>:

hammad13060:
$$ is even better, this even makes the syntax much simpler.

djmitche:
Using a fixed variable name would preclude nested maps, right?  It would also be less clear (especially using the very Perl-like `$$`).

I do like the `each(<identifier>)` form: it is very readable.

As for variable shadowing, I think that's a small cost to users.  Maps will typically be only a few lines long, and the scope of the variable is those few lines.  If they use `each(pullRequest)` and that shadows a global `pullRequest` variable that they use in the map, well, it should be pretty clear from inspection what went wrong.

hammad13060:
Hmm now that you mentioned about nested maps, I think using a fixed variable will have some issues in such cases.

hammad13060:
@jonasfj 

Can you add some cases where not including ```by``` key in case of **object arrays** should throw an error.

jonasfj:
@hammad13060, there is no `by` key is this PR... but we probably show throw if map isn't given an array.

hammad13060:
This looks fine now 👍

PR #9, opened 2017-01-22
Proposing $sort and $reverse

jonasfj:
With this proposal `by(identifier)` is optional for `$sort`, but necessary if sorting an object.

Generally, I suspect these are useful.

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -184,3 +184,41 @@ context:  {a: 3, b: 2}
 template: {$eval: '"" + (a + b)'}
 result:   '5'
 ################################################################################
+---
+section:  $sort
+---
+title:    simple sort
+context:  {}
+template: {$sort: [3, 4, 1, 2]}
+result:   [1, 2, 3, 4]
+---
+title:    simple sort with $eval
+context:  {array: [3, 4, 1, 2]}
+template: {$sort: {$eval: 'array'}}
+result:   [1, 2, 3, 4]
+---
+title:    sort by
+context:  {}
+template:
+  $sort: [{a: 2}, {a: 1, b: []}, {a: 3}]
+  by(x): 'x.a'
+result:   [{a: 1, b: [], {a: 2}, {a: 3}}]
+################################################################################
+---
+section:  $reverse
+---
+title:    simple reverse
+context:  {}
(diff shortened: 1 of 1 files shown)


specification.yml
```diff
@@ -184,3 +184,41 @@ context:  {a: 3, b: 2}
...
+title:    simple sort with $eval
+context:  {array: [3, 4, 1, 2]}
+template: {$sort: {$eval: 'array'}}
+result:   [1, 2, 3, 4]
+---
+title:    sort by
+context:  {}
+template:
+  $sort: [{a: 2}, {a: 1, b: []}, {a: 3}]
+  by(x): 'x.a'
```
hammad13060: Instead of specifying chained expression lets just specify the object property name we want to compare on

i.e
```yaml
by: 'a'
```

hammad13060: Just for clarification ```by``` can have a chained expression, however its scope should be descendants of objects to be compared.

hammad13060:
@jonasfj 
Can you add some cases where not including by key in case of object arrays should throw an error.

specification.yml
jonasfj: if you alllow `by: 'a'`, then if you have a list object structures and way to do: `by: 'key.array[2]'`, then you'll have to invent a third expression language. Yeah, you would want the same syntax, but the initial rule of the grammar would be different. Hence, a different language.

Also you wouldn't be able to do: `by(x): '-x.k'`, or `by(x): 'lowercase(x.k)'` (assuming the context has a function for converting strings to lower case. or `by(x): 'min(2, x.k)'` or `by(x): 'min(x.k, x.y)'`...

hammad13060: Ahh I see ... now it makes sense to me :)

jonasfj: Perhaps we should add an example with `by(x): `min(x.a, x.b)`

hammad13060: We have not introduced specs for function evaluation anywhere, so I think we should file this in a separate PR.

hammad13060: okay I just realised that in test conditions you have mentioned that context must have functions min, max at the time of testing so I think you cann add the above example in this PR.

PR #12, opened 2017-01-22
expressions for array indexing and array slicing

hammad13060:
I hope this should be enough.

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -244,10 +244,65 @@ context: {a: [1,2,3,4]}
 template: {$eval: 'a[0]'}
 result: 1
 ---
+title: array access [index]
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[2]'}
+result: 3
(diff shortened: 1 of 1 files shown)


specification.yml
```diff
@@ -238,6 +238,16 @@ title:    $eval must take string
...
+---
+title: array access [index]
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[0]'}
+result: 1
+---
+title: array slicing [begin:end], [begin, end)
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[0:2]'}
+result: [1,2]
```
jonasfj: I think we should support `a[:2]` as well, and `a[2:]`...
Probably also `a[-1] == 3` (ie. counting from end of the array, like in python..).

That way we can do nifty things like `a[-3:]` to get the last 3 elements of an array.

PR #15, opened 2017-02-02
Pratt-Interpreter Integration

hammad13060:
This PR contains ```code/tests``` for ```pratt-interpreter```. I will file separate PR for json-e tests once all its features will be implemented.


hammad13060:
@jonasfj 

I added support for array slicing. Checkout examples [here](https://github.com/taskcluster/json-e/pull/15/files#diff-5b72c4d94ad22a3b7e8c94bfaf38f31dR131)

I am not able to figure out how to implement slicing like ```[begin:]``` or ```[:end]```. Any suggestions ?

package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "json-e",
+  "version": "1.0.0",
+  "description": "json parameterization module inspired from json-parameterization",
+  "main": "./dist/index.js",
+  "scripts": {
+    "compile": "babel-compile -p es2015 src:lib test:.test",
```
jonasfj: See https://github.com/taskcluster/taskcluster-lib-monitor/blob/master/package.json#L8
Use taskcluster preset, that way we have `async/await` which we might want...

src/index.js
```diff
@@ -0,0 +1,181 @@
+var interpreter = require('./interpreter');
+var _ = require('lodash');
+var moment = require('moment');
+
+var safeEval = (src, context) => interpreter.parse(src, context);
```
jonasfj: Always prefer `let` to `var` that goes everywhere...

src/index.js
```diff
@@ -0,0 +1,181 @@
...
+
+module.exports = function(_template, _context) {
```
jonasfj: I propose we refactor this as a `render` function that given a template returns a value by cloning or transformation. That way we can call `render` recursively, example:
```js
// Do the string interpolation stuff
let interpolate = (string, context) => {
  return string.replace(/\${([^}]*)}/g, (text, expr) => interpreter.parse(expr, context));
};

let deleteMarker = {}; // value used to indicate a delete marker

let constructs = {
  '$if': (template, context) => {
    if (interpreter.parse(template['$if'], context)) {
      return template.then ? render(template.then, context) : deleteMarker;
    }
    return template.else ? render(template.else, context) : deleteMarker;
  },
  ....
}

let render = (template, context) => {
  if (typeof template === 'number' || typeof template === 'boolean') {
    return template;
  }
  if (typeof template === 'string') {
    return interpolate(template, context);
  }
  if (template instanceof Array) {
    return template.map(value => render(value, context)).filter(value !== deleteMarker);
  }

  // TODO: Probably throw an error if template two construct keys like having both $if and $map,
  for (let construct of Object.keys(constructs)) {
    if (template.hasOwnKey(construct)) {
      return constructs[construct](template, context);
    }
  }

  // clone object
  let result = {};
  for (let key of Object.keys(template)) {
    let value = render(template[key], context);
    if (value !== deleteMarker) {
      result[interpolate(key, context)] = value;
    }
  }
  return result;
};

module.exports = render;
```

This is just an outline, but the idea is that through recursive calls we clone the entire template...
And by defining constructs as a mapping from `$<name>` to function we can easily check that all the constructs are there... This outline lacks a lot of sanity checks...

src/interpreter.js
```diff
@@ -0,0 +1,106 @@
...
+    },
+    '.': (left, token, ctx) => left[ctx.require('id').value],
+    '(': (left, token, ctx) => left.apply(null, parseList(ctx, ',', ')')),
+    '==': (left, token, ctx) => left === ctx.parse('=='),
+    '!=': (left, token, ctx) => left !== ctx.parse('!='),
+    '<': (left, token, ctx) => left < ctx.parse('<'),
+    '<=': (left, token, ctx) => left <= ctx.parse('<='),
+    '>': (left, token, ctx) => left > ctx.parse('>'),
+    '>=': (left, token, ctx) => left >= ctx.parse('>='),
+    ':': (left, token, ctx) => [left, ctx.parse()],
```
jonasfj: What expression is this? This is not the way to handle `variable[1:4]`

src/interpreter.js
```diff
@@ -0,0 +1,106 @@
...
+    '{':      (token, ctx) => parseObject(ctx),
+    string: (token, ctx) => _.trim(_.trim(token.value, '\''), '"'),
+  },
+  infixRules: {
+    '+': (left, token, ctx) => left + ctx.parse('+'),
+    '-': (left, token, ctx) => left - ctx.parse('-'),
+    '*': (left, token, ctx) => left * ctx.parse('*'),
+    '/': (left, token, ctx) => left / ctx.parse('/'),
+    '[': (left, token, ctx) => {
+      let v = ctx.parse();
```
jonasfj: So they key to implementing this is:
```js
(left, token, ctx) => {
  let a, b;
  if (ctx.try(':')) {
    a = 0;
    if (ctx.try(']') {
      b = left.length;
      ctx.require(']');
    }
  } else {
    a = ctx.parse();
    if (ctx.try(':') {
      if (ctx.try(']') {
        b = left.length;
      } else {
        b = ctx.parse();
        ctx.require(']');
      }
    }
  }
  if (a !== undefined && b !== undefined) {
    return left.slice(a, b)
  }
  return left[a];
},
```

This is still missing sanity checks... for example, if it is an `left[a:b]` expression then `left` **must** be an array, we should check that before we do `left.length` to ensure that we don't get the left property of an object.. and to ensure that we throw a nice error message if the user is trying to do range index on an object :)

jonasfj: Basically, `ctx.try(':')` tries to consume a `':'`, but if it can't it returns `null` and doesn't consume anything... So we can use it in an if-condition...

Note: This kind of parsing is essentially recursive-decent parsing... If the separator was `+` instead of `:` we wold have a problem, or we would have to specify the correct precedence level when we do the recursive parse step...
Lucky for us, `:` and `]` aren't used for other expressions... and `]` is always balanced with `[`...

jonasfj: Note: I would move this out into a function like `parseInterval`, just like I did with `parseList` and `parseObject` these things are bound to get complicated.

src/interpreter.js
jonasfj: See comment above..

src/interpreter.js
```diff
@@ -0,0 +1,106 @@
...
+      ctx.require(']'); return v;
+    },
+    '.': (left, token, ctx) => left[ctx.require('id').value],
+    '(': (left, token, ctx) => left.apply(null, parseList(ctx, ',', ')')),
+    '==': (left, token, ctx) => left === ctx.parse('=='),
+    '!=': (left, token, ctx) => left !== ctx.parse('!='),
+    '<': (left, token, ctx) => left < ctx.parse('<'),
+    '<=': (left, token, ctx) => left <= ctx.parse('<='),
+    '>': (left, token, ctx) => left > ctx.parse('>'),
+    '>=': (left, token, ctx) => left >= ctx.parse('>='),
```
jonasfj: I like this... May I request that we align the `(left, token, ctx) =>` expressions... That makes it easier to write..

Also let's add a function for these... I'm thinking:
```js
{
  ...,
  '>=': (left, token, ctx) => compareNumbers(left, token.value, ctx.parse('>='))
}
```
That way we can have `compareNumbers` throw an error if either `left` or `right` isn't a number...
Good error messages are great :)

Note: In fact we can probably do:
```js
{
  patterns: {
    comparison: '>=|<=|>|<'
  },
  infixRules: {
    'comparison': (left, token, ctx) => compareNumbers(left, token.value, ctx.parse('comparison')),
  }
}
```

Then we only need one rule... Note: `compareNumbers` has the signature: `compareNumbers(left, op, right)` example: `compareNumbers(23, '>=', 45)`

src/prattparser.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return token;
+  }
+
+  parse(precedence = null) {
+    precedence = precedence === null ? 0 : this._precedenceMap[precedence];
+    let token = this.require();
+    let prefixRule = this._prefixRules[token.kind];
+    if (!prefixRule) {
+      throw new Error('...');
```
jonasfj: We obivously have to write something in these errors...

I suggest that we make a custom error subclass, see:
https://github.com/taskcluster/taskcluster-verify-cot/blob/8bf6baf3a75a968177915b38a9e54fd75e4c02be/src/proofbuilder.js#L85-L92

With properties such as: `start` and `end` index in the source, `message`...
Notably, the `start` and `end` is important because we can catch the error and display the message on the relevant line in  codemirror, if we're building an interactive editor :)

Notice that all tokens have `start` and `end` properties so it should be easy to add something like that every time we throw an error... Probably we just need a constructor `new SyntaxError(message, start, end)`.

Similarly, we'll probably need an `new InterpreterError(message, start, end)` for throw type errors, etc.. from our interpreter...

jonasfj: Note: I think both parser and tokenizer can use the `SyntaxError` but interpret might need a different error class, since we can't call those syntax errors :)

jonasfj:
This is a really good start though! :)

package.json
```diff
@@ -0,0 +1,39 @@
...
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hammad13060/json-e.git"
+  },
+  "author": "Hammad Akhtar <hammad13060@iiitd.ac.in>",
+  "license": "MIT",
+  "dependencies": {
+    "assume": "^1.4.1",
+    "lodash": "^4.17.2",
```
hammad13060: I am using it elsewhere apart from tests.

jonasfj: I noticed, but I think we should aim to avoid it.

Basically, it's large dependency for relatively small things.. Given that this will likely be used in many contexts I think we should try to avoid any unnecessary dependencies.
We want a python implementation too, using dependencies will just make it harder to make those two implementations match up.

Note: Normally I'm all for lodash, this probably the exception rather then the rule.

hammad13060:
Hi @jonasfj 

I have refactored most of the code. Still some work regarding ```throwing Interpreter errors and syntax``` is left. I am thinking we can finish all json-e features mentioned in specs within this PR only with tests passing.

Have a look :)

src/index.js
```diff
@@ -0,0 +1,94 @@
...
+    if (value !== deleteMarker) {
+      result[interpolate(key, context)] = value;
+    }
+  }
+  return result;
+};
+
+console.log(render({a: {$eval: '1 + 2'}}, {}));
+
+module.exports = render;
```
jonasfj: I propose that we wrap `render` before we export it, so that it can't ever return `deleteMarker` :)

```js
  let result = render(template, context);
  if (result === deleteMarker) {
    return undefined;
  }
  return result;
};
```

I don't know if `deleteMarker` should evaluate as an error or as `undefined`, I suspect `undefined` or `null` is ideal...
Using `undefined` is almost the same as `null` except it can be distinguished if anybody would ever want to.

src/interpreter.js
```diff
@@ -0,0 +1,141 @@
...
+      this.require(':');
+      let v = this.parse();
+      obj[k] = v;
+    } while (this.try(','));
+    this.require('}');
+  }
+  return obj;
+};
+
+let parseKey = (left, token, ctx) => {
```
jonasfj: let's call it `parseRange` as it's a range `[a:b]`, or `parseInterval`...

src/interpreter.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+};
+
+let compareNumbers = (left, operator, right) => {
+  switch (operator) {
+    case '>=': return left >= right;
+    case '<=': return left <= right;
+    case '>': return left > right;
+    case '<': return left < right;
+    default: return left < right;
```
jonasfj: Let's use default to throw a syntax error for unknown operator...

src/interpreter.js
```diff
@@ -0,0 +1,141 @@
...
+  patterns: {
+    number: '[0-9]+(?:\\.[0-9]+)?',
+    id:     '[a-zA-Z_][a-zA-Z_0-9]*',
+    string: '\'[^\']*\'|"[^"]*"',
+    comparison: '>=|<=|>|<',
+  },
+  tokens: [
+    ...'+-*/[].(){}:,'.split(''),
+    'number', 'id', 'string',
+    '==', '!=', '===', '!==',
```
jonasfj: Note: the ordering here matters.

The tokenizer uses a regular expression and it always returns the first match.
Example:
```
> /==|===/.exec('===')
[ '==', index: 0, input: '===' ]
> /===|==/.exec('===')
[ '===', index: 0, input: '===' ]
```
So if we have two tokens or patterns that could be the same, we have to list the longest one first. Or the preferred one first... In the case of `+` and `-` a string can't match both so it doesn't matter.
In the case of `if` and `id` we would have to list `if` before `id` or every time it saw the string `'if'` it would assume it's an identifier.

So let's put `number`, `id` and `string` at the very end of this list...


------
Also let's not add `===` and `!==` to our expression language... We can implement `==` as `===`, since nobody should ever want `4 == '4'` anyways.

hammad13060:
@jonasfj 

Have a look at this.

src/prattparser.js
```diff
@@ -88,14 +99,16 @@ class Context {
...
+      throw InterpreterError('No prefix rule of kind: ' + token.kind + ', value: ' + token.value,
+        token.start, token.end);
     }
     let left = prefixRule(token, this);
     while (this._next && precedence < this._precedenceMap[this._next.kind]) {
       let token = this.require();
       let infixRule = this._infixRules[token.kind];
       if (!infixRule) {
-        throw new Error('...');
+        throw InterpreterError('No infix rule of kind: ' + token.kind + ', value: ' + token.value,
```
jonasfj: The more I look at this... I think we should change the while condition so that this can't happen. And then instead handle it in `PrattParser.parse` where it currently throws: `throw new Error('Expected end of input');`.

Instead of `throw new Error('Expected end of input');` we should probably make it something like:
```js
let token = ctx.try();
if (token) {
  throw new SyntaxError(`Missing operator, found '${token.value}' expected: ${Object.keys(this._infixRules).join(', ')}`, token);
}
```
So if I have `4 +5  7 * 3`
It'll tell me: `Missing operator, found '7' expected: +, -, *, /, [, ., (,`

src/prattparser.js
```diff
@@ -0,0 +1,120 @@
...
+           ignore: null,
+           patterns: {},
+           tokens: [],
+           precedence: [],
+           prefixRules: {},
+           infixRules: {},
+         }, options);
+
+    // TODO: Validate options
+    // Ensure we have precedence for all the kinds used in infixRules
```
jonasfj: This is probably smart to do... too...

src/prattparser.js
```diff
@@ -0,0 +1,120 @@
...
+
+    this._prefixRules = prefixRules;
+    this._infixRules = infixRules;
+  }
+
+  parse(source, context = {}, offset = 0) {
+    let ctx = new Context(this, source, context, offset);
+    let result = ctx.parse();
+    if (ctx.try()) {
+      throw new Error('Expected end of input');
```
jonasfj: Syntax error see comment further down... where I talk about this one.

hammad13060: Can you elaborate on this one. I don't see any comments further down in ```prattparser.js```.

hammad13060:
I have added implementation for all the constructs mentioned in ```specification.yml```. Probably $map and $sort implementation can be made more modular since both share some common code. For the time being have a look at the implementation of [$sort](https://github.com/taskcluster/json-e/pull/15/files#diff-1fdf421c05c1140f6d71444ea2b27638R101) and [$map](https://github.com/taskcluster/json-e/pull/15/files#diff-1fdf421c05c1140f6d71444ea2b27638R70). In case I am missing some cases let me know :)

src/index.js
```diff
@@ -74,30 +86,82 @@ let constructs = {
             let result = template['$map'].map((v) => {
               context[itr] = v;
```
jonasfj: Don't modify the context object, instead let's create a new context object, this reduces the risk of bugs...

```js
return render(template[prop], Object.assign({}, context, {[itr]: v}));
```

src/index.js
```diff
@@ -47,21 +55,25 @@ let constructs = {
...
   },
   $map: (template, context) => {
     let exp = /\(([a-zA-Z_][a-zA-Z_0-9]*)\)/;
-    if (template['$map'] instanceof Array) {
+    if (isObject(template['$map'])) {
+      template['$map'] = render(template['$map'], context);
+    }
+
+    if (isArray(template['$map'])) {
       for (let prop of Object.keys(template)) {
```
jonasfj: We have to refactor this:
 1. throw error if not array
 2. identifiy the `each(...)` key
 3. throw error if there is no `each(...)` key, or if there is additional keys
 4. do the map step
```js
let value = render(template.$map, context);
if (!isArray(value)) {
  throw ....
}

// Find the 'each(x)' property
if (Object.keys(template).length !== 2) {
  throw...
}
let eachKey = Object.keys(template).filter(k => k !== '$map')[0];
let match = /^each\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(eachKey);
if (!match) {
  throw...
}
let x = match[1];
let each = template[eachKey];

// do the map step
return value.map(v => render(each, Object.assign({}, context, {[x]: v})));
```

jonasfj: We should probably decide what identifiers look like... and:
A) Enforce that the initial `context` given to function we export doesn't define identifiers other than those.
B) Reuse that pattern for identifiers in our interpreter grammar...
C) Reuse that pattern for `each(...)` and `by(...)`

I suspect a pattern like: `[a-zA-Z_][a-zA-Z0-9_]*` will be decent, thoughts?

src/index.js
```diff
@@ -74,30 +86,82 @@ let constructs = {
...
+    }
+
+    if (isArray(template['$sort'])) {
+      let arr = template['$sort'];
+      if (isArray(arr[0]) || isObject(arr[0])) {
+        for (let prop of Object.keys(template)) {
+          if (template.hasOwnProperty(prop) && prop.startsWith('by')) {
+            let match = exp.exec(prop);
+            if (match && match[1]) {
+              let itr = match[1];
```
jonasfj: This suffers the same issue as `$map`, refactor similarly...

src/index.js
```diff
@@ -74,30 +86,82 @@ let constructs = {
...
+        return template['$sort'].sort();
       }
+
+    } else {
+      jsonTemplateError('$sort requires array as value\n');
+    }
+  },
+  $reverse: (template, context) => {
+    if (isObject(template['$reverse'])) {
+      template['$reverse'] = render(template['$reverse'], context);
```
jonasfj: Don't modify template...

src/index.js
```diff
@@ -74,30 +86,82 @@ let constructs = {
...
+      } else {
+        return template['$sort'].sort();
       }
+
+    } else {
+      jsonTemplateError('$sort requires array as value\n');
+    }
+  },
+  $reverse: (template, context) => {
+    if (isObject(template['$reverse'])) {
```
jonasfj: This is never conditional...

src/index.js
jonasfj: In short:
```js
let value = render(template.$reverse, context);
if (!isArray(value)) {
  throw jsonTemplateError('$reverse expected an array template', template);
}
return value.reverse();
```

src/interpreter.js
```diff
@@ -65,6 +65,9 @@ let parseInterval = (left, token, ctx) => {
     if (begin !== undefined && end !== undefined) {
       return left.slice(begin, end);
     }
+    if (begin < 0) {
+      begin = (left.length + begin) % left.length;
+    }
     return left[begin];
   } else {
```
jonasfj: Avoid else, prefer early returns... Oh, and probably add comments...

src/interpreter.js
```diff
@@ -65,6 +65,9 @@ let parseInterval = (left, token, ctx) => {
     if (begin !== undefined && end !== undefined) {
       return left.slice(begin, end);
     }
+    if (begin < 0) {
+      begin = (left.length + begin) % left.length;
```
jonasfj: Super nice...

src/interpreter.js
```diff
@@ -0,0 +1,154 @@
...
+      let v = this.parse();
+      obj[k] = v;
+    } while (this.try(','));
+    this.require('}');
+  }
+  return obj;
+};
+
+let parseInterval = (left, token, ctx) => {
+  if (left instanceof Array) {
```
jonasfj: I think this is unwise, it's going to yield value sensitive parsing...
It's probably best to parse... Then later throw an error if `left` was expected to be an array...

src/interpreter.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+};
+
+let compareNumbers = (left, operator, right) => {
+  switch (operator) {
+    case '>=': return left >= right;
+    case '<=': return left <= right;
+    case '>': return left > right;
+    case '<': return left < right;
+    default: throw InterpreterError('no rule for comparison operator: ' + operator);
```
jonasfj: `new Error('Unknown comparison operator')`,

This should be an interpreter error... this is an internal error... This should never be able to happen.

src/interpreter.js
```diff
@@ -0,0 +1,154 @@
...
+};
+
+module.exports = new PrattParser({
+  ignore: '\\s+', // ignore all whitespace including \n
+  patterns: {
+    number: '[0-9]+(?:\\.[0-9]+)?',
+    id:     '[a-zA-Z_][a-zA-Z_0-9]*',
+    string: '\'[^\']*\'|"[^"]*"',
+    comparison: '>=|<=|>|<',
+    bool: 'true|false',
```
jonasfj: As previously suggested I think we should avoid making `comparison` and `bool` patterns... As it'll produce better error messages if they are tokens... I know it adds a few more rules... but these should be very simple rules.

src/interpreter.js
```diff
@@ -0,0 +1,154 @@
...
+  prefixRules: {
+    number: (token, ctx) => Number(token.value),
+    '-':      (token, ctx) => - ctx.parse('unary'),
+    '+':      (token, ctx) => ctx.parse('unary'),
+    id:     (token, ctx) => ctx.context[token.value],
+    '[':      (token, ctx) => ctx.parseList(',', ']'),
+    '(':      (token, ctx) => {let v = ctx.parse(); ctx.require(')'); return v;},
+    '{':      (token, ctx) => parseObject(ctx),
+    string: (token, ctx) => token.value.slice(1, -1),
+    bool: (token, ctx) => token.value === 'true' ? true : false,
```
jonasfj: Let's try to align these... please use spaces instead of tabs...

src/prattparser.js
```diff
@@ -0,0 +1,124 @@
...
+    this._infixRules = parser._infixRules;
+    this._next = this._tokenizer.next(this._source, offset);
+    this.context = context;
+  }
+
+  /**
+   * Try to get the next token if it matches one of the kinds given, otherwise
+   * return null. If no kinds are given returns the next of any kind.
+   */
+  try(...kinds) {
```
jonasfj: Randomly, let's rename this method to `attempt`, it's a small thing...
But I was playing with python and noticed that `try` is a reserved keyword, and we can't make methods called `try` in python...

It's a small change, but will allow us to keep the code and concepts the same when we port it to python... (not in this PR)

src/prattparser.js
```diff
@@ -0,0 +1,124 @@
...
+    let prefixRule = this._prefixRules[token.kind];
+    if (!prefixRule) {
+      syntaxRuleError(token, this._infixRules);
+    }
+    let left = prefixRule(token, this);
+    while (this._next && precedence < this._precedenceMap[this._next.kind]) {
+      let token = this.require();
+      let infixRule = this._infixRules[token.kind];
+      if (!infixRule) {
+        syntaxRuleError(token, this._prefixRules);
```
jonasfj: Change the `while` condition so that this can't happen... 
and throw this error at `throw new Error('Expected end of input');`....

src/prattparser.js
jonasfj: https://github.com/taskcluster/json-e/pull/15#discussion_r99431972

src/index.js
```diff
@@ -47,21 +55,25 @@ let constructs = {
...
-      throw new TemplateError('$switch can evaluate string expressions only\n' + JSON.stringify(template, null, '\t'));
+    if (!isString(template['$switch'])) {
+      jsonTemplateError('$switch can evaluate string expressions only\n');
     }
     let c = interpreter.parse(template['$switch'], context);
     return template.hasOwnProperty(c) ? render(template[c], context) : deleteMarker;
   },
   $json: (template, context) => {
-    if (template['$json'] instanceof Array || template['$json'] instanceof Object) {
+    if (isArray(template['$json']) || isObject(['$json'])) {
```
hammad13060: Ya seems right :)

src/index.js
hammad13060: WOW! This makes logic look super easy and readable :)

hammad13060: ```[a-zA-Z_][a-zA-Z0-9_]*``` covers most possible range of identifiers. I think we can go for this.

hammad13060:
Have a look :)

src/interpreter.js
```diff
@@ -0,0 +1,159 @@
...
+    default:   throw new Error('no rule for comparison operator: ' + operator);
+  }
+};
+
+module.exports = new PrattParser({
+  ignore: '\\s+', // ignore all whitespace including \n
+  patterns: {
+    number: '[0-9]+(?:\\.[0-9]+)?',
+    id:     '[a-zA-Z_][a-zA-Z_0-9]*',
+    string: '\'[^\']*\'|"[^"]*"',
```
jonasfj: I'm okay with this for a start... let's file an issue to figure out what string representation we want...
Notably, I'm thinking we probably want to support things this `\n`, etc... and unicode `\uXXXX` or whatever the escape sequences are...

I suggest we investigate what js, c++, python does and find some balance between these options... But let's discuss string escape sequences in a separate issue. Please file one :)

src/interpreter.js
```diff
@@ -0,0 +1,159 @@
...
+      let v = this.parse();
+      obj[k] = v;
+    } while (this.attempt(','));
+    this.require('}');
+  }
+  return obj;
+};
+
+let parseInterval = (left, token, ctx) => {
+  let begin, end;
```
jonasfj: Let's defined the variables: `let a = null, b = null, isInterval = false;`
Then we leave set `isInterval = true` if we read a `:` and set `a` and `b` as given, leaving them `null` if we have `[:]`...

That we can bother with `left` and concerns about the type of `left` later...

src/interpreter.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+    ctx.attempt(']');
+    return left.slice(begin, end);
+  }
+  // support for -ve index access
+  if (typeof begin === 'number' && begin < 0) {
+    begin = (left.length + begin) % left.length;
+  }
+  ctx.attempt(']');
+  return left[begin];
```
jonasfj: We probably need to do `hasOwnProperty` and if it's an array, we probably have to forbid access to anything that isn't an integer...
ie. people should not be allowed to do `([1,2,3])['indexOf']` because `indexOf` is a javascript function...
Similarly we want to avoid things like `({})['prototype']` hehe :) 

Worst case these could be used to get hold of global objects like `window` or `globals` in node.js and then you can get hold of `eval` and effectively escape the sandbox... fun right...

jonasfj: For starters separating the parsing and access step... Then it'll be a lot easier to reason about security aspects later...

jonasfj: We probably also have to handle crazy cases like: `myObj[{'2': 45}]`, obviously, using `{'2': 45}` to index both arrays and objects should fail with an error...

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+  } 
+  if (isString(template)) {
+    return interpolate(template, context);
+  }
+  if (isArray(template)) {
+    return template.map((v) => render(v, context)).filter((v) => v !== deleteMarker);
+  }
+
+  // check for multiple constructs
+  let detectConstruct = false, prevConstruct;
```
jonasfj: `prevConstruct` is never used

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+  for (let construct of Object.keys(constructs)) {
+    if (template.hasOwnProperty(construct)) {
+      if (detectConstruct) {
+        throw new TemplateError('only one construct allowed\n'
+          + JSON.stringify(template, null, '\t'));
+      } else {
```
jonasfj: else not necessary...

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+    return interpolate(template, context);
+  }
+  if (isArray(template)) {
+    return template.map((v) => render(v, context)).filter((v) => v !== deleteMarker);
+  }
+
+  // check for multiple constructs
+  let detectConstruct = false, prevConstruct;
+  for (let construct of Object.keys(constructs)) {
+    if (template.hasOwnProperty(construct)) {
```
jonasfj: Instead of all this... perhaps we do:
```js
let matches = Object.keys(constructs).filter(c => template.hasOwnProperty(c));
if (matches.length > 1) {
  throw....
}

if (matches.length === 1) {
  return constructs[matches[0]](template, context);
}
```

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+    return value.map(v => render(each, Object.assign({}, context, {[x]: v}))).filter(v => v !== deleteMarker);
+
+  },
+  $sort: (template, context) => {
+    let value = render(template['$sort'], context);
+    if (!isArray(value)) {
+      throw jsonTemplateError('$sort requires array as value\n', template);
+    }
+
+    if (isArray(value[0]) || isObject(value[0])) {
```
jonasfj: Why does this matter.... I can use `'by(x)':  'min(x + 1, 10)'` if I want to...

Generally if you're doing `isXXXX` on something returned from `render` for any purposes other than throwing an error, you're likely making a mistake.

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+      if (Object.keys(template).length !== 2) {
+        throw jsonTemplateError('$sort requires cannot have more than two properties\n', template);
+      }
+
+      let byKey = Object.keys(template).filter(k => k !== '$sort')[0];
+      let match = /^by\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(byKey);
+      if (!match) {
+        throw jsonTemplateError('$sort requires by(identifier) syntax\n', template);
```
jonasfj: `$sort` doesn't require this... but if it doesn't have `by(x)` then all elements in the array must be primitives...

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+    let eachKey = Object.keys(template).filter(k => k !== '$map')[0];
+    let match = /^each\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(eachKey);
+    if (!match) {
+      throw jsonTemplateError('$map requires each(identifier) syntax\n', template);
+    }
+
+    let x = match[1];
+    let each = template[eachKey];
+
+    return value.map(v => render(each, Object.assign({}, context, {[x]: v}))).filter(v => v !== deleteMarker);
```
jonasfj: NIT: Can we break `.filter` to next line?

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+    let c = interpreter.parse(template['$switch'], context);
+    return template.hasOwnProperty(c) ? render(template[c], context) : deleteMarker;
+  },
+  $json: (template, context) => {
+    return JSON.stringify(render(template['$json'], context));
+  },
+  $reverse: (template, context) => {
+    let value = render(template['$reverse'], context);
+    if (!isArray(value)) {
+      throw jsonTemplateError('$reverse requires array as value\n', template);
```
jonasfj: let's say `$reverse requires an array template` or `$reverse requires a template for an array`...
The error wasn't that `$reverse` wasn't given an array value, the error was that the **template** that `$reverse` was given didn't evaluated to an array, when it was rendered.

src/index.js
```diff
@@ -0,0 +1,186 @@
...
+      throw new TemplateError('Cannot interpolate objects/arrays: '
+        + text + ' <-- ' + expr);
+    }
+    return v.toString();
+  });
+};
+
+let deleteMarker = {};
+
+let constructs = {
```
jonasfj: So this is a minor proposal to reduce indentation level... given that many of these constructs aren't small functions, especially not when we start working on making good error messages (later).
Perhaps we should it as:
```js
let constructs = {};

// {$if: '4 == 5', then: 'yes', else: 'no'} -> 'no'
constructs.$if = (template, context) => {
  ...
};

// {$fromNow: '.......
```

Just so that we have room for a few comments, that reminds us what we're doing in a given construct.

package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "files": [
+    "lib"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hammad13060/json-e.git"
+  },
+  "author": "Hammad Akhtar <hammad13060@iiitd.ac.in>",
+  "license": "Mozilla Public License Version 2.0",
```
jonasfj: `MPL-2.0`

See: https://docs.npmjs.com/files/package.json#license

There is actually a list of license identifiers here: https://spdx.org/licenses/

src/interpreter.js
```diff
@@ -0,0 +1,159 @@
...
+    string: '\'[^\']*\'|"[^"]*"',
+  },
+  tokens: [
+    ...'+-*/[].(){}:,'.split(''),
+    'number', 'true', 'false', 'id', 'string',
+    '>=', '<=', '<', '>', '==', '!=',
+  ],
+  precedence: [
+  	['==', '!='],
+  	['>=', '<=', '<', '>'],
```
hammad13060: Have a look at [python operator precedence](https://docs.python.org/dev/reference/expressions.html#operator-precedence) and [js operator precedence](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table)

src/interpreter.js
```diff
@@ -0,0 +1,159 @@
...
+  
+  prefixRules: {
+    number: (token, ctx) => Number(token.value),
+    '-':    (token, ctx) => - ctx.parse('unary'),
+    '+':    (token, ctx) => ctx.parse('unary'),
+    id:     (token, ctx) => ctx.context[token.value],
+    '[':    (token, ctx) => ctx.parseList(',', ']'),
+    '(':    (token, ctx) => {let v = ctx.parse(); ctx.require(')'); return v;},
+    '{':    (token, ctx) => parseObject(ctx),
+    string: (token, ctx) => token.value.slice(1, -1),
```
hammad13060: I thnik this should work. What do you say @jonasfj ?
```javascript
if (str[0] === '"') {
  return str.replace('"', '\"').slice(1, -1);
}
return str.replace('\"', '"').slice(1, -1);
```

jonasfj: Yeah, that looks good... We can add more escape sequences later..
I'm not sure we need them.. but I guess they could be nice to have...

package.json
```diff
@@ -14,10 +14,10 @@
   ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/hammad13060/json-e.git"
+    "url": "https://github.com/taskcluster/json-e"
   },
-  "author": "Hammad Akhtar <hammad13060@iiitd.ac.in>",
```
jonasfj: feel free to keep your name... or if really fancy figure out how the new alternative to `authors` works... I think `authors` have been deprecated, but I never bothered the read about the new field...

src/index.js
```diff
@@ -31,104 +31,112 @@ let interpolate = (string, context) => {
...
+  let value = render(template['$sort'], context);
+  if (!isArray(value)) {
+    throw jsonTemplateError('$sort requires array as value\n', template);
+  }
+
+  let byKey = Object.keys(template).filter(k => k !== '$sort')[0];
+  let match = /^by\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(byKey);
+  if (!match) {
+    let needBy = value.some(v => isArray(v) || isObject(v)).length;
+    if (needBy) {
```
jonasfj: The `.length` is probably a left over... `value.some()` returns a boolean...
`if (value.some(v => isArray(v) || isObject(v)) { ` should do

src/index.js
```diff
@@ -31,104 +31,112 @@ let interpolate = (string, context) => {
...
+
+  let x = match[1];
+  let by = template[byKey];
+  let contextClone = Object.assign({}, context);
+
+  return value.sort((left, right) => {
+    contextClone[x] = left;
+    left = render(by, contextClone);
+    contextClone[x] = right;
+    right = render(by, contextClone);
```
jonasfj: This isn't a `render` isn't `by(x): 'max(x.mykey, 10)'` supposed to be an expression.... so use `interpreter.parse`...

Randomly I'm tempted to suggest that we should instead do:
```js
return value.map(v => {
  return {v, k: interpreter.parse(by, Object.assign({}, context, {[x]: v}))};
}).sort((a, b) => {
  if (a.k <= b.k) { // I didn't double check if this was sane
    return false;
  }
  return true;
}).map(({v}) => v);
```

So that way we don't  compute `by` of any entry in `value` more than once... I'm thinking `interpreter.parse` is probably relatively slow...

src/index.js
```diff
@@ -0,0 +1,184 @@
...
+
+  let matches = Object.keys(constructs).filter(c => template.hasOwnProperty(c));
+  if (matches.length > 1) {
+    throw jsonTemplateError('only one construct allowed\n', template);
+  }
+  if (matches.length === 1) {
+    return constructs[matches[0]](template, context);
+  }
+
+  for (let construct of Object.keys(constructs)) {
```
jonasfj: shouldn't be needed anymore, right?

src/index.js
```diff
@@ -0,0 +1,184 @@
...
+    let value = render(template[key], context);
+    if (value !== deleteMarker) {
+      result[interpolate(key, context)] = value;
+    }
+  }
+  return result;
+};
+
+module.exports = (template, context = {}) => {
+  let result = render(template, context);
```
jonasfj: Let's do an assert that `Object.keys()` always match `/[a-zA-Z_][a-zA-Z0-9_]*/` otherwise they can't be referenced from the interpreter, right?

src/interpreter.js
jonasfj: Ah, yeah, this looks right...

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+    return str.replace('"', '\"').slice(1, -1);
+  }
+  return str.replace('\"', '"').slice(1, -1);
+};
+
+module.exports = new PrattParser({
+  ignore: '\\s+', // ignore all whitespace including \n
+  patterns: {
+    number: '[0-9]+(?:\\.[0-9]+)?',
+    id:     '[a-zA-Z_][a-zA-Z_0-9]*',
```
jonasfj: Let's change the name to `identifier` then it'll look better in error messages...

src/index.js
```diff
@@ -0,0 +1,184 @@
...
+    let v = interpreter.parse(expr, context);
+    if (v instanceof Array || v instanceof Object) {
+      throw new TemplateError('Cannot interpolate objects/arrays: '
+        + text + ' <-- ' + expr);
+    }
+    return v.toString();
+  });
+};
+
+let deleteMarker = {};
```
jonasfj: Please give this a comment... `// Object used to indicate deleteMarker

src/index.js
```diff
@@ -0,0 +1,184 @@
...
+let isString = (expr) => typeof expr === 'string';
+let isNumber = (expr) => typeof expr === 'number';
+let isBool = (expr) => typeof expr === 'boolean';
+let isArray = (expr) => expr instanceof Array;
+let isObject = (expr) => expr instanceof Object;
+
+let jsonTemplateError = (msg, template) => new TemplateError(msg + JSON.stringify(template, null, '\t'));
+
+let interpolate = (string, context) => {
+  return string.replace(/\${([^}]*)}/g, (text, expr) => {
```
jonasfj: We probably should support `${myfunction({'key': 23})}`, but this won't because the regex will match: `${myfunction({'key': 23}`, we can't do this with regular expressions...

So yeah, this gets **fun**... let's file a bug for it, and leave a TODO here...
I think we have to search for `${` and then run the interpreter, and not throw an unexpected token error if next token is `}` after parsing is done.... That'll probably require some tweaks... Let's file a an issue.

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+    true:   (token, context) => true,
+    false:  (token, context) => false,
+  },
+  infixRules: {
+    '+':        (left, token, ctx) => left + ctx.parse('+'),
+    '-':        (left, token, ctx) => left - ctx.parse('-'),
+    '*':        (left, token, ctx) => left * ctx.parse('*'),
+    '/':        (left, token, ctx) => left / ctx.parse('/'),
+    '[':        (left, token, ctx) => parseInterval(left, token, ctx),
+    '.':        (left, token, ctx) => left[ctx.require('id').value],
```
jonasfj: We probably need some safety here... hasOwnProperty, and isObject, and `!isArray`.

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+    ['*', '/'],
+    ['[', '.'],
+    ['('],
+    ['unary'],
+  ],
+  prefixRules: {
+    number: (token, ctx) => Number(token.value),
+    '-':    (token, ctx) => - ctx.parse('unary'),
+    '+':    (token, ctx) => ctx.parse('unary'),
+    id:     (token, ctx) => ctx.context[token.value],
```
jonasfj: Again we probably have to do some `hasOwnProperty`

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+    '{':    (token, ctx) => parseObject(ctx),
+    string: (token, ctx) => parseString(token.value),
+    true:   (token, context) => true,
+    false:  (token, context) => false,
+  },
+  infixRules: {
+    '+':        (left, token, ctx) => left + ctx.parse('+'),
+    '-':        (left, token, ctx) => left - ctx.parse('-'),
+    '*':        (left, token, ctx) => left * ctx.parse('*'),
+    '/':        (left, token, ctx) => left / ctx.parse('/'),
```
jonasfj: For `-*/` we probably have to check that we're operating on numbers...
For `+` we should be operating on numbers or strings...

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  if (a instanceof Object) {
+    throw new InterpreterError('cannot use objects as keys');
+  }
+  return obj[a];
+};
+
+let compareNumbers = (left, operator, right) => {
+  switch (operator) {
```
jonasfj: Let's check that `left` and `right` are numbers... So we don't compare arrays, objects, booleans this way...

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+  return accessProperty(left, a, b, isInterval);
+};
+
+let accessProperty = (obj, a, b, isInterval) => {
+  if (obj instanceof Array) {
+    if (isInterval) {
+      if (typeof a !== 'number' || typeof b !== 'number') {
+        throw new InterpreterError('cannot perform interval access with non-integers');
+      }
+      return obj.slice(a, b);
```
jonasfj: Don't we have to do stuff like `a = a < 0 ? (obj.length + a) % obj.length : a;` for both a and b, here too?

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+    a = a < 0 ? (obj.length + a) % obj.length : a;
+    return obj[a];   
+  }
+
+  // if we reach here it means we are accessing property value from object
+  if (!obj.hasOwnProperty(a)) {
+    throw new InterpreterError(`'${a}' not found in ${JSON.stringify(obj, null, '\t')}`);
+  }
+
+  if (a instanceof Object) {
```
jonasfj: We should probably insist that a is a string.... rather that not an object,.,..

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+      }
+      if (ctx.attempt(']')) {
+        b = left.length;
+      } else {
+        b = ctx.parse();
+      }
+    }
+  }
+
+  ctx.attempt(']');
```
jonasfj: I think this is required, right?

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+  return obj;
+};
+
+let parseInterval = (left, token, ctx) => {
+  let a = null, b = null, isInterval = false;
+  if (ctx.attempt(':')) {
+    a = 0;
+    isInterval = true;
+    if (!(left instanceof Array)) {
```
jonasfj: Leave this test to `accessProperty`

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+      throw new InterpreterError('cannot perform interval access on non-array');
+    }
+    if (ctx.attempt(']')) {
+      b = left.length;
```
jonasfj: Leave it `null` then do this in `accessProperty`

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      b = ctx.parse();
```
jonasfj: After this you need to require `]`

src/interpreter.js
jonasfj: it's probably the wrong place to do it...

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+    a = ctx.parse();
+    if (ctx.attempt(':')) {
+      isInterval = true;
+      if (!(left instanceof Array)) {
+        throw new InterpreterError('cannot perform interval access on non-array');
+      }
+      if (ctx.attempt(']')) {
+        b = left.length;
+      } else {
+        b = ctx.parse();
```
jonasfj: Again here you need to require ]

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+      }
+    }
```
jonasfj: else `require(]`

src/interpreter.js
```diff
@@ -0,0 +1,167 @@
...
+};
+
+let parseObject = (ctx) => {
+  let obj = {};
+  if (!this.attempt('}')) {
+    do {
+      let k = this.require('id', 'string');
+      this.require(':');
+      let v = this.parse();
+      obj[k] = v;
```
jonasfj: don't we have to do `k.value` ?

jonasfj: maybe make sure we have a few test cases :)

test/tests.js
```diff
@@ -0,0 +1,70 @@
...
+      jsoneSuite[currentSection].push(doc);
+    }
+  });
+
+  for (let currentSuite of Object.keys(jsoneSuite)) {
+    if (jsoneSuite.hasOwnProperty(currentSuite)) {
+      suite(currentSuite, () => {
+        jsoneSuite[currentSuite].forEach((c) => {
+          test(c.title, () => {
+            tk.freeze(date);
```
jonasfj: Neat trick...  but let's do it in before() and after() hooks...  for the entire suite... then we never forget...
```
suite('...', () => {
  before(() => tk.freeze(date));
  after(() => tk.freeze(date));

  test(...., () => {...});
});
```

test/tests.js
```diff
@@ -0,0 +1,70 @@
...
+    if (doc.hasOwnProperty('section')) {
+      jsoneSuite[doc.section] = [];
+      currentSection = doc.section;
+    } else {
+      jsoneSuite[currentSection].push(doc);
+    }
+  });
+
+  for (let currentSuite of Object.keys(jsoneSuite)) {
+    if (jsoneSuite.hasOwnProperty(currentSuite)) {
```
jonasfj: Why do you need this?

src/index.js
hammad13060: forgot to remove this. hehehe :p

test/tests.js
hammad13060: probably got into the habit of checking for hasOwnProperty :p

src/interpreter.js
hammad13060: no need since .```.slice(begin, end)``` works for -ve indices as well :)

src/interpreter.js
hammad13060: I think we can define infix rules and prefix rules like this

```javascript
let infixRules = {}

infixRules['+'] = (left, token, ctx) => {.....}

module.exports = new PrattParser({
  .....,
  infixRules: infixRules,
  prefixRules: prefixRules,
  .....,
});
```

jonasfj: yeah,. that might make sense... we probably have to give up all hope of making them all one-liners :)

src/interpreter.js
hammad13060: probably need to throw some errors in case object syntax is not valid. Same for parseList.

src/interpreter.js
hammad13060: I think we can compare strings as well.

src/interpreter.js
hammad13060: I am wrong above. Parser will take care of errors :)

src/index.js
hammad13060: yes this version will be relatively slower since we will clone the context again and again. I think we should stick with older version with the bug fixed.

hammad13060: and yes I should also remove the else statement.

jonasfj:
I'm going to close an reopen this PR... I really tempted to land it (normally I would).

But since we're going to use this for configuration, breaking compatibility won't be good.. hence, we need quite a few review cycles... I'll probably file some PRs too, so it's not all on you :)

PR #18, opened 2017-02-06
prefix, infix rules updated, $sort bug fixed, separate module for js …

hammad13060:
filed this new PR because previous one got too overwhelmed :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -159,8 +159,6 @@ let render = (template, context) => {
   // clone object
   let result = {};
   for (let key of Object.keys(template)) {
-    assert(/[a-zA-Z_][a-zA-Z0-9_]*/.exec(key)[0], 
-      jsonTemplateError('keys must follow /[a-zA-Z_][a-zA-Z0-9_]*/\n', template).msg);
     let value = render(template[key], context);
     if (value !== deleteMarker) {
       result[interpolate(key, context)] = value;
@@ -169,28 +167,9 @@ let render = (template, context) => {
   return result;
 };
 
-let assertContext = (context) => {
-
-  if (isNumber(context) || isBool(context) || isString(context) || isFunction(context)) {
-    return;
-  }
-
-  if (isArray(context)) {
-    context.forEach(v => assertContext(v));
-    return;
-  }
-
-  for (let key of Object.keys(context)) {
-    if (context.hasOwnProperty(key)) {
-      assert(/[a-zA-Z_][a-zA-Z0-9_]*/.exec(key)[0], 
-      jsonTemplateError('context keys must follow /[a-zA-Z_][a-zA-Z0-9_]*/\n', context).msg);
-      assertContext(context[key]);
-    }
-  }
-};
-
 module.exports = (template, context = {}) => {
-  assertContext(context);
+  Object.keys(context).forEach(v => assert(/[a-zA-Z_][a-zA-Z0-9_]*/.exec(v)[0], 
+    jsonTemplateError('top level keys must follow /[a-zA-Z_][a-zA-Z0-9_]*/\n', template).msg));
   let result = render(template, context);
   if (result === deleteMarker) {
     return undefined;
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,10 @@
 let interpreter = require('./interpreter');
 let fromNow = require('./from-now');
 let ExtendableError = require('es6-error');
+let assert = require('assert');
+
+let {isString, isNumber, isBool, 
+  isArray, isObject, isFunction} = require('./type-utils');
 
 class TemplateError extends ExtendableError {
   constructor(message) {
@@ -9,12 +13,6 @@ class TemplateError extends ExtendableError {
     this.name = 'TemplateError';
   }
 }
-
-let isString = (expr) => typeof expr === 'string';
-let isNumber = (expr) => typeof expr === 'number';
-let isBool = (expr) => typeof expr === 'boolean';
-let isArray = (expr) => expr instanceof Array;
-let isObject = (expr) => expr instanceof Object;
 
 let jsonTemplateError = (msg, template) => new TemplateError(msg + JSON.stringify(template, null, '\t'));
 
@@ -29,6 +27,7 @@ let interpolate = (string, context) => {
   });
 };
 
+// Object used to indicate deleteMarker
 let deleteMarker = {};
 
 let constructs = {};
@@ -115,7 +114,7 @@ constructs.$sort = (template, context) => {
   let byKey = Object.keys(template).filter(k => k !== '$sort')[0];
   let match = /^by\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(byKey);
   if (!match) {
-    let needBy = value.some(v => isArray(v) || isObject(v)).length;
+    let needBy = value.some(v => isArray(v) || isObject(v));
     if (needBy) {
       throw jsonTemplateError('$sort requires by(identifier) for sorting arrays of objects/arrays\n', template);
     }
@@ -128,14 +127,13 @@ constructs.$sort = (template, context) => {
 
   return value.sort((left, right) => {
     contextClone[x] = left;
-    left = render(by, contextClone);
+    left = interpreter.parse(by, contextClone);
     contextClone[x] = right;
-    right = render(by, contextClone);
+    right = interpreter.parse(by, contextClone);
     if (left <= right) {
       return false;
-    } else if (left > right) {
-      return true;
     }
+    return true;
   });
 };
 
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -158,15 +156,11 @@ let render = (template, context) => {
...
-  for (let construct of Object.keys(constructs)) {
-    if (template.hasOwnProperty(construct)) {
-      return constructs[construct](template, context);
-    }
-  }
-
   // clone object
   let result = {};
   for (let key of Object.keys(template)) {
+    assert(/[a-zA-Z_][a-zA-Z0-9_]*/.exec(key)[0],
```
jonasfj: I see no such restriction?

src/index.js
```diff
@@ -175,7 +169,28 @@ let render = (template, context) => {
   return result;
 };
 
+let assertContext = (context) => {
+
+  if (isNumber(context) || isBool(context) || isString(context) || isFunction(context)) {
+    return;
```
jonasfj: I don't think this allowed...

src/index.js
```diff
@@ -175,7 +169,28 @@ let render = (template, context) => {
...
+  if (isArray(context)) {
+    context.forEach(v => assertContext(v));
+    return;
+  }
+
+  for (let key of Object.keys(context)) {
+    if (context.hasOwnProperty(key)) {
+      assert(/[a-zA-Z_][a-zA-Z0-9_]*/.exec(key)[0], 
+      jsonTemplateError('context keys must follow /[a-zA-Z_][a-zA-Z0-9_]*/\n', context).msg);
+      assertContext(context[key]);
```
jonasfj: I don't think we need recursion...
If I have context: `{myVar: {'hello-world': 5}}` I can reference it using: `myVar['hello-world']` 
I can't do: `myVar.hello-world` that won't work...
If I had context:  `{'hello-world': 5}` then I would be in trouble, because  there is no way to do: `global['hello-world']`, because we haven't defined a `global` variable... So we should just forbid top-level keys in context from being invalid identifiers....

src/interpreter.js
```diff
@@ -26,87 +30,87 @@ let parseList = (ctx, separator, terminator) => {
 
 let parseObject = (ctx) => {
   let obj = {};
-  if (!this.attempt('}')) {
+  if (!ctx.attempt('}')) {
```
jonasfj: nice catch! :)

src/interpreter.js
```diff
@@ -26,87 +30,87 @@ let parseList = (ctx, separator, terminator) => {
...
 let parseInterval = (left, token, ctx) => {
-  let a = null, b = null, isInterval = false;
+  let a = null, isInterval = false;
   if (ctx.attempt(':')) {
     a = 0;
     isInterval = true;
-    if (!(left instanceof Array)) {
-      throw new InterpreterError('cannot perform interval access on non-array');
-    }
-    if (ctx.attempt(']')) {
```
jonasfj: no.... Don't remove this....
Do parsing first... then interpretation later...

Basically, do not access `left` and `right` in parseInterval, and when you get to `accessProperty` don't access `ctx`...

jonasfj: Splitting the code in a parse-step and a evaluation-step makes it simpler... It means you won't throw context sensitive syntax errors...

src/interpreter.js
```diff
@@ -26,87 +30,87 @@ let parseList = (ctx, separator, terminator) => {
...
 };
 
 let compareNumbers = (left, operator, right) => {
-  switch (operator) {
-    case '>=': return left >= right;
-    case '<=': return left <= right;
-    case '>':  return left > right;
-    case '<':  return left < right;
-    default:   throw new Error('no rule for comparison operator: ' + operator);
+  let valid = isNumber(left) && isNumber(right) || isString(left) && isString(right);
```
jonasfj: Prefer early return...
```js
if (!(isNumber(left) && isNumber(right) || isString(left) && isString(right))) {
  throw ...
}
```
then you reduce indentation level...

src/interpreter.js
```diff
@@ -116,16 +120,140 @@ let parseString = (str) => {
   return str.replace('\"', '"').slice(1, -1);
 };
 
+let testOperand = (operator, operand) => {
+  if (operator === '+') {
```
jonasfj: This is as simple as:
```js
if (!isNumber(operand) && (operator !== '+' || !isString(operand))) {
  throw ....
}
```

src/interpreter.js
```diff
@@ -116,16 +120,140 @@ let parseString = (str) => {
...
+prefixRules['false'] = (token, ctx) => {
+  if (token.value === 'false') {
+    return false;
+  }
+  throw new Error('Only \'true/false\' is considered as bool');
+};
+
+// infix rule definition starts here
+infixRules['+'] = infixRules['-'] = infixRules['*'] = infixRules['/']
+  = (left, token, ctx) => applyMathOperator(left, token, ctx);
```
jonasfj: Nice, let's just move applyMathOperator down here... no need to have a separate method...

src/interpreter.js
```diff
@@ -116,16 +120,140 @@ let parseString = (str) => {
...
+infixRules['('] =  (left, token, ctx) => {
+  if (isFunction(left)) {
+    return left.apply(null, parseList(ctx, ',', ')'));
+  }
+  throw expectationError('infix: f(args)', 'f to be function');
+};
+
+infixRules['=='] = (left, token, ctx) => left === ctx.parse('==');
+infixRules['!='] = (left, token, ctx) => left !== ctx.parse('!=');
+infixRules['<='] = (left, token, ctx) => compareNumbers(left, token.value, ctx.parse('<='));
```
jonasfj: you can do the same as you did for applyMathOperator

src/interpreter.js
```diff
@@ -116,16 +120,140 @@ let parseString = (str) => {
...
-    string: '\'[^\']*\'|"[^"]*"',
+    number:     '[0-9]+(?:\\.[0-9]+)?',
+    identifier: '[a-zA-Z_][a-zA-Z_0-9]*',
+    string:     '\'[^\']*\'|"[^"]*"',
   },
   tokens: [
     ...'+-*/[].(){}:,'.split(''),
-    'number', 'true', 'false', 'id', 'string',
+    'number', 'true', 'false', 'identifier', 'string',
     '>=', '<=', '<', '>', '==', '!=',
```
jonasfj: Please list these before identifier, just for good measure... number, identifier, string should be last.
Notice that if `identifier` was listed before `true` then 'true' would always be an identifier, so we don't want that... 

So to avoid that, let's try to keep pattern tokens listed last.

src/interpreter.js
```diff
@@ -137,31 +265,6 @@ module.exports = new PrattParser({
...
-    '(':        (left, token, ctx) => left.apply(null, parseList(ctx, ',', ')')),
-    '==':       (left, token, ctx) => left === ctx.parse('=='),
-    '!=':       (left, token, ctx) => left !== ctx.parse('!='),
-    '<=':       (left, token, ctx) => compareNumbers(left, token.value, ctx.parse('<=')),
-    '>=':       (left, token, ctx) => compareNumbers(left, token.value, ctx.parse('>=')),
-    '<':        (left, token, ctx) => compareNumbers(left, token.value, ctx.parse('<')),
-    '>':        (left, token, ctx) => compareNumbers(left, token.value, ctx.parse('>')),
-  },
+  prefixRules: prefixRules,
+  infixRules: infixRules,
```
jonasfj: you can just do: `{prefixRules}`, it's the same as `{prefixRules: prefixRules}`, when variable and key is the same..

src/interpreter.js
hammad13060: I have done it slightly differently have a look.

src/index.js
hammad13060: good point :)

src/index.js
```diff
@@ -176,6 +168,8 @@ let render = (template, context) => {
 };
 
 module.exports = (template, context = {}) => {
+  Object.keys(context).forEach(v => assert(/[a-zA-Z_][a-zA-Z0-9_]*/.exec(v)[0], 
+    jsonTemplateError('top level keys must follow /[a-zA-Z_][a-zA-Z0-9_]*/\n', template).msg));
```
jonasfj: not sure what the msg is for

jonasfj: just do an assert using `.some`

src/interpreter.js
```diff
@@ -43,70 +47,59 @@ let parseInterval = (left, token, ctx) => {
...
-        b = left.length;
-      } else {
-        b = ctx.parse();
-      }
     }
   }
 
-  ctx.attempt(']');
+  if (isInterval && ctx.attempt(']')) {
+    b = left.length;
```
jonasfj: Use `null` as indicator that length should be inferred from `left.length`, then handle the type errors later...

hammad13060:
Have a look :)

PR #20, opened 2017-02-07
json-e progress

jonasfj:
Let's work on json-e in a branch, so we can have a PR against empty master... This allows us to make comments across everything... 

Adding bstack and dustin for review... They'll probably get around to it next week.


jonasfj:
All my review comments are in PR #15, they've pretty much been addressed.

Please note:
Correctness and backwards compatibility is key for json-e. So anything you spot that can improve this please comment on it.

Imagine we use json-e for `.taskcluster.yml` in hg and git repositories, if we then change the behavior of json-e in a non-backwards compatible manner, then we risk breaking all those files. That would be bad.
Hence, an behavior that currently work, kind of has to keep working, when we publish this.

For that reason we have to double check that things like `[].indexOf` isn't accessible, etc.. That we don't allow `2 + '2'` and other stupid behavior that we would want to fix later. Because fixing it later could cause breakages, and hence, be extremely hard to roll out.

jonasfj:
Oh, for the record @hammad13060 wrote most of this :)

specification.yml
```diff
@@ -186,16 +194,31 @@ context:  {}
 template: {$fromNow: '1 m'}
 result:   '2017-01-19T16:28:20.974Z'
 ---
+title:    fromNow 1m
+context:  {}
+template: {$fromNow: '1m'}
+result:   '2017-01-19T16:28:20.974Z'
```
imbstack: What's the plan for testing these timestamps? Since this is kinda a set of test cases, right? Do we set the "Now" to be a certain time and then run the tests or update all of these `results` fields first?

imbstack: Oh, oops. I saw how it was done in the test file later and forgot to delete this. All is well.

src/interpreter.js
```diff
@@ -0,0 +1,301 @@
...
+  if (isFunction(left)) {
+    return left.apply(null, parseList(ctx, ',', ')'));
+  }
+  throw expectationError('infix: f(args)', 'f to be function');
+};
+
+infixRules['=='] = infixRules['!='] = infixRules['<='] = 
+infixRules['>='] = infixRules['<'] =  infixRules['>'] 
+  =  (left, token, ctx) => {
+    let operator = token.value;
```
jonasfj: let's not do this...
For all these tokens we have that `token.value == token.kind`, but in practice those are two different concepts..

`token.value` is the string value that made up the token. `token.kind` is the kind of token...
You definitely want to be using `kind` here not  `value`...

Keeping the distinction is good for sanity, we have some tokens where we use a pattern, and then `value` is not the same as `kind`... And who knows what tokens we might introduce in the future...

src/interpreter.js
```diff
@@ -0,0 +1,301 @@
...
+  tokens: [
+    ...'+-*/[].(){}:,'.split(''),
+    '>=', '<=', '<', '>', '==', '!=', '!', '&&', '||',
+    'true', 'false', 'number', 'identifier', 'string',
+  ],
+  precedence: [
+    ['||'],
+    ['&&'],
+  	['==', '!='],
+  	['>=', '<=', '<', '>'],
```
jonasfj: huh, indentation ?

src/interpreter.js
```diff
@@ -0,0 +1,301 @@
...
+};
+
+let parseString = (str) => {
+  if (str[0] === '"') {
+    return str.replace('"', '\"').slice(1, -1);
+  }
+  return str.replace('\"', '"').slice(1, -1);
+};
+
+let tetsComparisonOperands = (operator, left, right) => {
```
jonasfj: Name :)

src/interpreter.js
hammad13060: fixing in #34

src/interpreter.js
hammad13060: Yes, I agree. Nice point. Fixing in #34 :)

src/interpreter.js
hammad13060: Fixing in #34 👍

hammad13060:
@jonasfj 
Is it looking good enough for landing ?

src/prattparser.js
```diff
@@ -0,0 +1,131 @@
...
+    this._prefixRules = prefixRules;
+    this._infixRules = infixRules;
+  }
+
+  parse(source, context = {}, offset = 0) {
+    let ctx = new Context(this, source, context, offset);
+    let result = ctx.parse();
+    let next = ctx.attempt();
+    if (next) {
+      throw syntaxRuleError(next, Object.keys(this._infixRules).join(', '));
```
hammad13060: @jonasfj 
I think this error can occur because of following reasons:
1. We wanted to use some infix character but ended up using something like ```}, ], etc``` (Error message covers this case)
2. Someone added an extra ```}, ]``` at the end of source hence we got a stray token back when we did ctx.attempt(), hence we should also show ```Expected End of Input``` error.

What say ?

jonasfj: You can also get it if someone adds an extra `3` example: `3  3` or `(4 +2) 3`

jonasfj: But you right the error message isn't good for the expected end of input case...

Maybe we want the message to say, we expected one of these... or end of input?... Or what?

djmitche: If this is still an open question, please file an issue for it

jonasfj:
yeah, merging is fine... I'll put filing more issues in my todo list :)

--

2017-03-13 9:13 GMT-07:00 Dustin J. Mitchell <notifications@github.com>:

PR #25, opened 2017-02-09
Invalid Expression Tests

hammad13060:
some tests for invalid expressions :)


test/expression-language.yml
```diff
@@ -396,4 +426,109 @@ result: {a: 1, b: 1}
...
+context: {key: 'hello'}
+source: '+key'
+error: true
+---
+title: 'invalid expression (3)'
+context: {key: 'hello'}
+source: 'key2'
+error: true
+---
+title: 'invalid expression (4)'
```
jonasfj: isn't this the same as case (3)

Note. a better title is `'undefined variable'`

test/expression-language.yml
```diff
@@ -396,4 +426,109 @@ result: {a: 1, b: 1}
...
+error: true
+---
+title: 'invalid expression (1)'
+context: {key: 'hello'}
+source: '-key'
+error: true
+---
+title: 'invalid expression (2)'
+context: {key: 'hello'}
+source: '+key'
```
jonasfj: "Prefix + type error" is a better title

test/expression-language.yml
```diff
@@ -158,6 +163,31 @@ context: {key: [1,2,3,4,5]}
 source: 'key[:-3]'
 result: [1,2]
 ---
+title: 'array slicing (7)'
```
jonasfj: `array slicing type error` is a better title

test/expression-language.yml
```diff
@@ -396,4 +426,109 @@ result: {a: 1, b: 1}
...
+context: {}
+source: '"hello" <= 3'
+error: true
+---
+title: 'invalid expression (17)'
+context: {}
+source: '"hello" >= 3'
+error: true
+---
+title: 'invalid expression (18)'
```
jonasfj: "Comparison type error" is a better title for these...

test/expression-language.yml
```diff
@@ -396,4 +426,109 @@ result: {a: 1, b: 1}
...
+error: true
+---
+title: 'invalid expression (7)'
+context: {}
+source: 'true + false'
+error: true
+---
+title: 'invalid expression (8)'
+context: {}
+source: '"hello" - 3'
```
jonasfj: out of curiosity do we allow: `'hello' + 3` ? should we?

Ie. int + string...

hammad13060: yes we do.

hammad13060: I just think we should disallow this feature because in python we will get a ```TypeError``` for doing so. What say ?

PR #26, opened 2017-02-10
boolean operators

hammad13060:
proposal for ```logical and```, ```logical or``` and ```logical not```.

### DIFF
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -131,6 +131,12 @@ let testMathOperand = (operator, operand) => {
   return;
 };
 
+let testLogicalOperand = (operator, operand) => {
+  if (!isBool(operand)) {
+    throw expectationError(`infix: ${operator}`, `boolean ${operator} boolean`);
+  }
+};
+
 let prefixRules = {};
 let infixRules = {};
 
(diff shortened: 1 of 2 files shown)


test/expression-language.yml
```diff
@@ -28,6 +28,66 @@ context: {a: 3.0, b: 2.0}
 source: 'b / a'
 result: 0.6666666666666666
 ---
+title: 'logical not (1)'
+context: {key1: true, key2: false}
```
jonasfj: I would rename the context `{tt: true, ff: false}`
That way the truth table becomes a lot easier to read...

src/interpreter.js
```diff
@@ -131,6 +131,12 @@ let testMathOperand = (operator, operand) => {
   return;
 };
 
+let testLogicalOperand = (operator, operand) => {
+  if (!isBool(operand)) {
+    throw expectationError(`infix: ${operator}`, `boolean ${operator} boolean`);
```
jonasfj: We probably should pass in `token` instead of `operator`...
Notably we want the position information from the `token` if we're going to throw an error... we probably should try to include position information in `InterpreterError`. But we can also do that in a separate PR.

Just a note.

jonasfj:
Feel free to hit merge.

src/interpreter.js
hammad13060: We should do this in a separate PR because we can use ```syntaxerror.js``` as a common error class. what do you say @jonasfj ?

Issue #19, opened 2017-02-06
Allow support for stuff like ${myfunction({'key': 23})}


hammad13060:
We probably should support ${myfunction({'key': 23})}, but this won't because the regex will match: ```${myfunction({'key': 23}```

PR #27, opened 2017-02-11
Interpolation algorithm patch

hammad13060:
New interpolation algorithm for [bug](https://github.com/taskcluster/json-e/issues/19) as suggested by @jonasfj :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,7 +22,7 @@ let interpolate = (string, context) => {
   let begin = 0;
   let remainingStr = string;
   while (true) {
-    let offset = remainingStr.search(/\${/);
+    let offset = remainingStr.search(/\${/g);
     if (offset === -1) {
       break;
     }
--- a/src/index.js
+++ b/src/index.js
@@ -18,14 +18,16 @@ class TemplateError extends ExtendableError {
 let jsonTemplateError = (msg, template) => new TemplateError(msg + JSON.stringify(template, null, '\t'));
 
 let interpolate = (string, context) => {
-  return string.replace(/\${([^}]*)}/g, (text, expr) => {
-    let v = interpreter.parse(expr, context);
-    if (v instanceof Array || v instanceof Object) {
-      throw new TemplateError('Cannot interpolate objects/arrays: '
-        + text + ' <-- ' + expr);
-    }
-    return v.toString();
+  let result = '';
+  let begin = 0;
+  string.replace(/\${/, (text, offset, str) => {
+    let v = interpreter.parseUntilTerminator(str.slice(offset), 2, '}', context);
+    let terminatorOffset = v.offset;
+    result += str.slice(begin, offset) + v.result.toString();
+    begin = terminatorOffset + offset + 1; 
   });
+  result += string.slice(begin);
+  return result;
 };
 
 // Object used to indicate deleteMarker
--- a/src/index.js
+++ b/src/index.js
@@ -20,13 +20,20 @@ let jsonTemplateError = (msg, template) => new TemplateError(msg + JSON.stringif
 let interpolate = (string, context) => {
   let result = '';
   let begin = 0;
-  string.replace(/\${/, (text, offset, str) => {
-    let v = interpreter.parseUntilTerminator(str.slice(offset), 2, '}', context);
-    let terminatorOffset = v.offset;
-    result += str.slice(begin, offset) + v.result.toString();
-    begin = terminatorOffset + offset + 1; 
-  });
-  result += string.slice(begin);
+  let remainingStr = string;
+  while (true) {
+    let offset = remainingStr.search(/\${/);
+    if (offset === -1) {
+      break;
+    }
+    let v = interpreter.parseUntilTerminator(remainingStr.slice(offset), 2, '}', context);
+    if (isArray(v.result) || isObject(v.result)) {
+      throw new TemplateError('cannot interpolate array/object: ' + string);
+    }
+    result += remainingStr.slice(0, offset) + v.result.toString();
+    remainingStr = remainingStr.slice(offset + v.offset + 1);
+  }
+  result += remainingStr;
   return result;
 };
 
--- a/src/index.js
+++ b/src/index.js
@@ -20,20 +20,17 @@ let jsonTemplateError = (msg, template) => new TemplateError(msg + JSON.stringif
 let interpolate = (string, context) => {
   let result = '';
   let begin = 0;
-  let remainingStr = string;
-  while (true) {
-    let offset = remainingStr.search(/\${/g);
-    if (offset === -1) {
-      break;
-    }
-    let v = interpreter.parseUntilTerminator(remainingStr.slice(offset), 2, '}', context);
+  let remaining = string;
+  let offset;
+  while ((offset = remaining.search(/\${/g)) !== -1) {
+    let v = interpreter.parseUntilTerminator(remaining.slice(offset), 2, '}', context);
     if (isArray(v.result) || isObject(v.result)) {
       throw new TemplateError('cannot interpolate array/object: ' + string);
     }
-    result += remainingStr.slice(0, offset) + v.result.toString();
-    remainingStr = remainingStr.slice(offset + v.offset + 1);
+    result += remaining.slice(0, offset) + v.result.toString();
+    remaining = remaining.slice(offset + v.offset + 1);
   }
-  result += remainingStr;
+  result += remaining;
   return result;
 };
 
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -249,7 +249,7 @@ infixRules['>='] = infixRules['<'] =  infixRules['>']
   };
 
 module.exports = new PrattParser({
-  ignore: '[\\s#\\$]+', // ignore all whitespace including \n
+  ignore: '\\s+', // ignore all whitespace including \n
   patterns: {
     number:     '[0-9]+(?:\\.[0-9]+)?',
     identifier: '[a-zA-Z_][a-zA-Z_0-9]*',
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -249,7 +249,7 @@ infixRules['>='] = infixRules['<'] =  infixRules['>']
   };
 
 module.exports = new PrattParser({
-  ignore: '\\s+', // ignore all whitespace including \n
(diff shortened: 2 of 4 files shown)


hammad13060:
I think we will have to reserve ```${``` for interpolation then only this will work, because if someone will ```${``` intentionally for non-interpolation purpose, code will raise errors.

src/index.js
```diff
@@ -18,14 +18,16 @@ class TemplateError extends ExtendableError {
...
-  return string.replace(/\${([^}]*)}/g, (text, expr) => {
-    let v = interpreter.parse(expr, context);
-    if (v instanceof Array || v instanceof Object) {
-      throw new TemplateError('Cannot interpolate objects/arrays: '
-        + text + ' <-- ' + expr);
-    }
-    return v.toString();
+  let result = '';
+  let begin = 0;
+  string.replace(/\${/, (text, offset, str) => {
```
jonasfj: I'm guessing something like `string.slice(offset).search(/\${/)` inside a `while(offset < string.length) {`..

Try adding a test cases like: `'My name is ${name} I am ${age} years old'` where you have two or more substitutions inside a string.

jonasfj:
>I think we will have to reserve ${ for interpolation then only this will work, because if someone will use ${ intentionally for non-interpolation purpose, code will raise errors.

As far as I can see options are:
 1. `${` always triggers an error if not used for interpolation
 2. If an interpolation triggers a parse error, we don't throw the error but just output the source.
    ie. `'${ 7 + 6 hello world'` -> `'${ 7 + 6 hello world'`
 3. if `${` isn't followed by a `}` anywhere in the string we don't interpolate.

I agree (1) is the sane strategy, because (2) and (3) are likely examples of errors that a user didn't want to do.

hammad13060:
@jonasfj 

current algorithm will fail if we will have something like ```${key1}%${key2}``` since interpreter doesn't know how to ```deal``` with ```%```. By ```deal``` I mean either ignore or tokenize ```%```. I think workaround will involve ignoring everything which is not specified in tokens array.

jonasfj:
No because the parser will never proceed past }
Instead it'll only interpolate the first ${...} instance... Try it...

hammad13060:
I tried and I get the following error 
screen shot 2017-02-14 at 2 15 14 pm

jonasfj:
I don't understand how does the interpreter ever get to `%`? It sees the next token as `}` and then we stop interpreting right?

```
abc${key1}%${key2}
   ^ first jump to here, right?
abc${key1}%${key2}
     ^ Then we start parsing from here
abc${key1}%${key2}
         ^ intepreter can't find an infix rule for }, so it returns
abc${key1}%${key2}
          ^ We continue searching for ${ from here
abc${key1}%${key2}
           ^ We find ${ here
abc${key1}%${key2}
             ^ we start parsing from here
```

hammad13060:
Yes this is what I think is the expected behaviour, but when ```ctx.attempt()``` is called something goes wrong and i.e. ```unexpected end of file error```

src/interpreter.js
```diff
@@ -249,7 +249,7 @@ infixRules['>='] = infixRules['<'] =  infixRules['>']
   };
 
 module.exports = new PrattParser({
-  ignore: '\\s+', // ignore all whitespace including \n
+  ignore: '[\\s#\\$]+', // ignore all whitespace including \n
```
jonasfj: Why ignore `#` and `$` ?

src/index.js
```diff
@@ -18,14 +18,23 @@ class TemplateError extends ExtendableError {
...
 let interpolate = (string, context) => {
-  return string.replace(/\${([^}]*)}/g, (text, expr) => {
-    let v = interpreter.parse(expr, context);
-    if (v instanceof Array || v instanceof Object) {
-      throw new TemplateError('Cannot interpolate objects/arrays: '
-        + text + ' <-- ' + expr);
+  let result = '';
+  let begin = 0;
+  let remainingStr = string;
+  while (true) {
```
jonasfj: `let offset = -1`

`while ((offset = remainingStr.search(...)) != -1) {`

src/prattparser.js
```diff
@@ -48,6 +48,16 @@ class PrattParser {
     }
     return result;
   }
+
+  parseUntilTerminator(source, offset, terminator, context) {
+    let ctx = new Context(this, source, context, offset);
+    let result = ctx.parse();
+    let next = ctx.attempt();
+    if (next.value !== terminator) {
```
jonasfj: should be `kind` not value.

jonasfj: both are the same for the token `}`

jonasfj:
Try to debug with some console.log, if you get the multiple interpolation working yet...
Or maybe try to run tokenizer on `${key1}%${key2}´ and see what the output is...

jonasfj:
> I tried and I get the following error

Oh, that errors looks the be from reaching EOF at parsing the last one... makes sense...

Perhaps we should change the tokenizer to return an `EOF` token... as a special case.
Obviously, we would need asserts in the constructor that people don't define `EOF` as a token kind, as it's reserved...
Then have `parse(...)` call `parseUntil(..., terminator = 'EOF',...)` as parsing is just a the case of parse until EOF... and throwing an error if you don't get the terminator...

If we take it one step further and modify the  `parse()` to return `{value, offset}`, then we can just let `parse()` have a parameter `terminator` that defaults to `EOF` if not specified... So we don't need two methods..

hammad13060:
I noticed that error is occurring because of when ```ctx.attempt()``` is called it first fetches the next token and stores it in a variable and then it updates the next variable so by calling tokenizer.next() function. Calling tokenizer.next() causes the error since now we are at ```%${b}```.

hammad13060:
Check this out :)

jonasfj:
Ah, that makes more sense... I might have been wrong about the EOF problem.

src/prattparser.js
```diff
@@ -77,6 +88,13 @@ class Context {
     return token;
   }
 
+  /*
+  * returns next token object
+  */
+  getNext() {
```
jonasfj: Okay, let's rename this `peek()` as you are peeking ahead without consuming anything.
We can also use this in place of `this._next` down in `parse`.

Note: I was trying to avoid too many methods on `Context`, but I think you're right we need this...
Or, we just have `parseUntilTerminator` read the `_next` variable...

The idea was that `_next` is a private variable, and hence, not part of the public interface... But given that we might need it in Parser it's nice to expose it as a public method... Let's call it `peek`.

Thinking more about this, I suspect some future rules might want `peek() == peek(1) == _next`, and `peek(2)` as looking two ahead, `peek(3)` for looking 3 ahead..
Notably, I think `peek(2)` would be necessary for implementing syntax for functions like `(a, b) => a + b`.
Also with `peek(2)` it would be easy to implement `attempt(token1, token2)` as attempting to take two at the same time, resulting in either getting both or getting none.. But we can consider that when we add more complicated logic in the future.


----
Hmm, as I continue to write I'm starting to think that maybe `attempt()` should not throw an error if the future token doesn't parse. Maybe we should save that error for later and throw it next time `attempt()` is called.

Something like:
```js
attempt(...) {
   if (this._next instanceof Error) {
     throw this._next;
   }
   ...

  try {
    this._next = this._tokenizer.next(this._source, token.end);
  } catch (err) {
    this._next = err;
  }
  return token;
}
```

jonasfj: That way we don't need `peek()` or `getNext()`

jonasfj:
@hammad13060,
It definitely seems like you found the underlying problem.
Let's either throw the error from attempt() later, as proposed at the end of my comment, or rename to peek().
I suspect that throwing the error later will result in better error messages and stack traces, as the error will get thrown from the rule that expected another token, and not from the rule that parsed the token before the bad token appeared.

hammad13060:
I really like the idea of throwing error ```later```. I have made the necessary changes, have a look at them and let me know :)

src/prattparser.js
```diff
@@ -48,6 +48,17 @@ class PrattParser {
     }
     return result;
   }
+
+  parseUntilTerminator(source, offset, terminator, context) {
+    let ctx = new Context(this, source, context, offset);
+    let result = ctx.parse();
+    console.log(source, result);
```
jonasfj: :)

src/prattparser.js
```diff
@@ -48,6 +48,17 @@ class PrattParser {
...
+    let next = ctx.attempt();
+    if (next.value !== terminator) {
```
jonasfj: `next.kind`... In theory why shouldn't I be allowed to parse until there is a `number` or some other token that matches a pattern :)

src/index.js
```diff
@@ -18,14 +18,23 @@ class TemplateError extends ExtendableError {
...
 
 let interpolate = (string, context) => {
-  return string.replace(/\${([^}]*)}/g, (text, expr) => {
-    let v = interpreter.parse(expr, context);
-    if (v instanceof Array || v instanceof Object) {
-      throw new TemplateError('Cannot interpolate objects/arrays: '
-        + text + ' <-- ' + expr);
+  let result = '';
+  let begin = 0;
+  let remainingStr = string;
```
jonasfj: could we rename this variable `remaining`

src/prattparser.js
```diff
@@ -67,13 +78,21 @@ class Context {
...
       return null;
     }
     if (kinds.length > 0 && kinds.indexOf(token.kind) === -1) {
       return null;
     }
-    this._next = this._tokenizer.next(this._source, token.end);
+    try {
+      this._next = this._tokenizer.next(this._source, token.end);
+    } catch (err) {
+      this._next = err;
```
jonasfj: I'm just realizing... that we probably ought to do: `this._nextErr = err`
and then do `if (this._nextErr) {throw this._nextErr;}`

Just for sanity... we should probably also set `this._nextErr = null` in the constructor...

I'm just realizing that:
 1. `instanceof` probably isn't very fast
 2. this causes node.js to deoptimize a function, and generally messes with all js optimization
 3. this probably bad practices...

src/index.js
jonasfj: Also this minor loop cleanup..

jonasfj:
@hammad13060, I like how simple we're able to keep it... It's shaping up to be pretty elegant...

hammad13060:
Ya it is looking great, thanks to your reviews :)

Issue #28, opened 2017-02-13
TODO: raise error when performing 'number + string'


hammad13060:
we should handle cases and raise error where user might try to something like ```"hello " + 5```

PR #29, opened 2017-02-14
raise error when performing 'number + string'

hammad13060:
patch for [bug](https://github.com/taskcluster/json-e/issues/28).


jonasfj:
@hammad13060  you can reference issues by typing # like #28 (infact github will provide auto-complete for that)...

Issue #33, opened 2017-02-20
Implement the in-operator from python


jonasfj:
We should implement the `in` operator from python... Like `3 in [1,3,4,6]`.
It's a very common thing to test for, and it's one less function we need in our standard library.

PR #34, opened 2017-02-20
in operator, exponentiation operator support and parseObject bug fix

hammad13060:
This PR is aimed at resolving #33, #34  and fixing bug which allows parsing for objects like ```{"a": 123, "b": 456}```.

### DIFF
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -109,7 +109,7 @@ let parseString = (str) => {
   return str.replace('\"', '"').slice(1, -1);
 };
 
-let tetsComparisonOperands = (operator, left, right) => {
+let testComparisonOperands = (operator, left, right) => {
 
   if (operator === '==' || operator === '!=') {
     return null;
@@ -265,7 +265,7 @@ infixRules['>='] = infixRules['<'] =  infixRules['>']
   =  (left, token, ctx) => {
     let operator = token.kind;
     let right = ctx.parse(operator);
-    tetsComparisonOperands(operator, left, right);
+    testComparisonOperands(operator, left, right);
     switch (operator) {
       case '>=': return left >= right;
       case '<=': return left <= right;
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -213,7 +213,11 @@ prefixRules['false'] = (token, ctx) => {
 infixRules['+'] = infixRules['-'] = infixRules['*'] = infixRules['/'] = infixRules['**']
   = (left, token, ctx) => {
     testMathOperand(token.value, left);
-    let right = ctx.parse(token.value);
+    let operator = token.value;
+    if (operator === '**') {
+      operator = '**-right-associative';
+    }
+    let right = ctx.parse(operator);
     testMathOperand(token.value, right);
     if (typeof left !== typeof right) {
       throw new InterpreterError(`TypeError: ${typeof left} ${token.value} ${typeof right}`);
@@ -321,6 +325,7 @@ module.exports = new PrattParser({
   	['>=', '<=', '<', '>'],
     ['+', '-'],
     ['*', '/'],
+    ['**-right-associative'],
     ['**'],
     ['[', '.'],
     ['('],
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -33,11 +33,11 @@ let parseObject = (ctx) => {
   if (!ctx.attempt('}')) {
     do {
       let k = ctx.require('identifier', 'string');
-      ctx.require(':');
-      let v = ctx.parse();
       if (k.kind === 'string') {
         k.value = parseString(k.value);
       }
+      ctx.require(':');
+      let v = ctx.parse();
       obj[k.value] = v;
     } while (ctx.attempt(','));
     ctx.require('}');
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -280,11 +280,22 @@ infixRules['||'] = infixRules['&&'] = (left, token, ctx) => {
 infixRules['in'] = (left, token, ctx) => {
   let right = ctx.parse('in');
   if (isObject(right) && !isArray(right)) {
+    if (isNumber(left)) {
+      throw expectationError('Infix: in', 'String query on Object');
+    }
     right = Object.keys(right);
   }
 
   if (!(isArray(right) || isString(right))) {
     throw expectationError('Infix: in', 'array/string/object to perform lookup');
+  }
+
+  if (isArray(right) && !(isNumber(left) || isString(left))) {
+    throw expectationError('Infix: in', 'String/Number query on Array');
+  }
+
+  if (isString(right) && !isString(left)) {
+    throw expectationError('Infix: in', 'String query on String');
   }
 
   return right.indexOf(left) !== -1;
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -35,6 +35,9 @@ let parseObject = (ctx) => {
       let k = ctx.require('identifier', 'string');
       ctx.require(':');
       let v = ctx.parse();
+      if (k.kind === 'string') {
+        k.value = parseString(k.value);
+      }
       obj[k.value] = v;
     } while (ctx.attempt(','));
     ctx.require('}');
@@ -273,6 +276,19 @@ infixRules['||'] = infixRules['&&'] = (left, token, ctx) => {
   }
 };
 
+infixRules['in'] = (left, token, ctx) => {
+  let right = ctx.parse('in');
+  if (isObject(right) && !isArray(right)) {
+    right = Object.keys(right);
+  }
+
+  if (!(isArray(right) || isString(right))) {
+    throw expectationError('Infix: in', 'array/string/object to perform lookup');
+  }
+
+  return right.indexOf(left) !== -1;
+};
+
 module.exports = new PrattParser({
   ignore: '\\s+', // ignore all whitespace including \n
   patterns: {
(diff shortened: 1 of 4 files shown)


src/interpreter.js
```diff
@@ -35,6 +35,9 @@ let parseObject = (ctx) => {
       let k = ctx.require('identifier', 'string');
       ctx.require(':');
       let v = ctx.parse();
+      if (k.kind === 'string') {
```
jonasfj: I suggest moving this up two lines... so that it's straight below `let k = ctx.require('identifier', 'string');`...

jonasfj: Also nice catch finding this bug..

test/expression-language.yml
```diff
@@ -530,12 +610,12 @@ error: true
 ---
 title: 'parse object (1)'
 context: {key: 1}
-source: '{a: key, b: key + 1}'
+source: '{a: key, "b": key + 1}'
```
jonasfj: Shouldn't we have a test case with `{'a': 45}` ?

test/expression-language.yml
```diff
@@ -368,6 +388,66 @@ context: {a: 3, b: 2}
 source: 'min(3, 2) + max(3, 2)'
 result: 5
 ---
+title: 'in operator on string, success (1)'
+context: {}
+source: '"abc" in "aabc"'
+result: true
```
jonasfj: What about `3 in 'a3b'` ? Do we want to support that, throw a type error or just return false...
I started some in PR https://github.com/taskcluster/json-e/pull/36
But I'm not sure I got them right...

Either we return false for all of them, or we throw an error... It's tempting to return false instead of throwing an error..
I mean  we don't want a type error for something like: `1 in [2, '3', 1]`, right?

But clearly `1 in [2, '3', 1]` is true, and `3 in [2, '3', 1]` is false... So is `{} in [2, '3', 1]` false or does it throw an error?
I'm almost starting to think that `[] in X` and `{} in X` throw an error... So number in string might also throw an error?, what about number in object? error too right?

src/interpreter.js
```diff
@@ -273,6 +277,19 @@ infixRules['||'] = infixRules['&&'] = (left, token, ctx) => {
...
+  let right = ctx.parse('in');
+  if (isObject(right) && !isArray(right)) {
+    right = Object.keys(right);
+  }
+
+  if (!(isArray(right) || isString(right))) {
+    throw expectationError('Infix: in', 'array/string/object to perform lookup');
+  }
+
+  return right.indexOf(left) !== -1;
```
jonasfj: As discussed in comment below... Should we throw an error if `left` if is not a string or number... Perhaps even only allow number if `right` is an array?

Or should we do deep equals for arrays and objects?

test/expression-language.yml
hammad13060: Since ```in``` operator is inspired from python, so it will only make sense that it works as close to python as possible

- If left is Number right can only be an array/object
- if left is string right must also be string
- if left is array/object we must throw error

jonasfj:
Hmm, but if left is a number and right is an object we always returns
false... Because unlike Python we only support string keys... Python can
have numbers as keys...

So maybe throw an error in that case too...IE if left is a number and right
is a string...

Den 20. feb. 2017 8.53 PM skrev "Hammad Akhtar" <notifications@github.com>:

jonasfj:
Meant if left is a number and right is an object.

Den 20. feb. 2017 9.03 PM skrev "Jonas Finnemann Jensen" <jopsen@gmail.com>:

hammad13060:
hmm agreed 👍

test/expression-language.yml
```diff
@@ -368,6 +388,91 @@ context: {a: 3, b: 2}
...
+---
+title: 'in operator on object, failure (6)'
+context: {}
+source: '3.4 in {"3": 1, "b": 2}'
+result: false
+---
+title: 'in operator on array, error (1)'
+context: {key: {a: 1, b: 2}}
+source: '[] in key'
+error: true
```
hammad13060: I have not added any code to raise errors this such case, still test passes this happens because of the following reason
- when source is evaluated it returns ```false``` (lets call this ```result````)
- ```c.result``` is ```undefined``` hence when we do ```assume(result).eql(c.result)```. assume statement raises an error and we land in catch clause and we never reach statement ```assert(!c.error, 'Expected an error from: ' + c.source);``` since ```return``` is called in catch statement.


I think this minor refactoring should work, what do you say
```javascript
    let result;
    try {
      result = interpreter.parse(c.source, _.defaults({}, c.context, builtinMethods));
    } catch (err) {
      if (!c.error) {
        throw err;
      }
      return;
    }
    assert(!c.error, 'Expected an error from: ' + c.source);
    assume(result).eql(c.result);
```

hammad13060: Just a thought
- I think that one should never do ```asserts```, ```assumes``` in ```try``` clause because if ```AssertionErrors``` will be raised we might land in ```catch``` clause and errors might never get thrown.

jonasfj: Ah, good point...

jonasfj: Feel free to refactor...

test/expression-language.yml
```diff
@@ -448,6 +453,26 @@ context: {key: {a: 1, b: 2}}
 source: '"c" in key'
 result: false
 ---
+title: 'in operator on object, failure (5)'
+context: {}
+source: '3 in {"3": 1, "b": 2}'
+result: false
```
jonasfj: Should we make this an error too?

hammad13060: yup

hammad13060:
Have a look

test/expression-language.yml
hammad13060: however something like this should work ```'"3" in {"3": 1, "b": 2}'.

src/interpreter.js
```diff
@@ -207,7 +210,7 @@ prefixRules['false'] = (token, ctx) => {
 };
 
 // infix rule definition starts here
-infixRules['+'] = infixRules['-'] = infixRules['*'] = infixRules['/']
+infixRules['+'] = infixRules['-'] = infixRules['*'] = infixRules['/'] = infixRules['**']
```
jonasfj: I suspect `'**'` is right associative...
See: https://en.wikipedia.org/wiki/Operator_associativity#Examples

ie `2 ** 2 ** 3 == 2 ** (2 ** 3)`, probably better add that as test case...
To get this we need to to add a new entry into the `precedence` array and use that in `ctx.parse(..)` instead of `ctx.parse('**')`.

So `precedence: [...., '**', '**-right-associative', ...] then use `ctx.parse('**-right-associative')`... Maybe a better name that what I used here...

hammad13060: WOW, nice trick 👍

hammad13060:
cannot think of a better name than ```**-right-associative``` right now :)

src/interpreter.js
```diff
@@ -213,7 +213,11 @@ prefixRules['false'] = (token, ctx) => {
 infixRules['+'] = infixRules['-'] = infixRules['*'] = infixRules['/'] = infixRules['**']
   = (left, token, ctx) => {
     testMathOperand(token.value, left);
-    let right = ctx.parse(token.value);
+    let operator = token.value;
+    if (operator === '**') {
+      operator = '**-right-associative';
+    }
+    let right = ctx.parse(operator);
```
jonasfj: Just make a different rule... No need to reuse the same logic...

jonasfj: ie. do `infixRules['**'] = (left, token, ctx) => {....`
Too many special cases for `**` isn't easy to read...

test/expression-language.yml
```diff
@@ -48,6 +48,26 @@ context: {a: 2, b: 3}
...
+---
+title: 'exponentiation, right associativity (3)'
+context: {a: 2, b: 3}
+source: '2 ** 2 ** 3 == 2 ** (2 ** 3)'
+result: true
+---
+title: 'exponentiation, right associativity (4)'
+context: {a: 2, b: 3}
+source: '2 ** 2 ** 3'
+result: 256
```
jonasfj: I like these :)

src/interpreter.js
```diff
@@ -209,21 +212,34 @@ prefixRules['false'] = (token, ctx) => {
...
-      case '+': return left + right;
-      case '-': return left - right;
-      case '*': return left * right;
-      case '/': return left / right;
+    switch (operator) {
+      case '+':  return left + right;
+      case '-':  return left - right;
+      case '*':  return left * right;
+      case '/':  return left / right;
+      case '**': return Math.pow(left, right);
```
jonasfj: One case too many

src/interpreter.js
```diff
@@ -209,21 +212,34 @@ prefixRules['false'] = (token, ctx) => {
...
       default: throw new Error(`unknown infix operator: '${operator}'`);
     }
   };
 
+infixRules['**'] = (left, token, ctx) => {
+  let operator = token.kind;
+  testMathOperand(operator, left);
+  let right = ctx.parse('**-right-associative');
+  testMathOperand(operator, right);
+  if (typeof left !== typeof right) {
```
jonasfj: What does this do?

hammad13060: This statement ensures that operators are applied on same types.

jonasfj:
But it operates on numbers only right? How can they not be given earlier
asserts...

Den 22. feb. 2017 10.17 AM skrev "Hammad Akhtar" <notifications@github.com>:

hammad13060:
Now I just realized that we don't need this statement here and we can refactor testMathOperand to ensure this

jonasfj:
Exactly...

Den 22. feb. 2017 10.23 AM skrev "Hammad Akhtar" <notifications@github.com>:

hammad13060:
I think this should do it :)

jonasfj:
feel free to land it, you have an r+

Issue #30, opened 2017-02-15
Define standard library (a base context with functions)


jonasfj:
@hammad13060,

I'm thinking that `json-e` should have some standard library... Well, a set of functions that a built-in.
What do you think?

My thinking is things like:
 * `sqrt(x)` for square root
 * `min/max(a,b)`
 * `str(x)` for converting to string
 * `uppercase(x)`
 * `lowercase(x)`
 * `abs(c)` absolute value
 * `hex(x)` to hexa decimal encoding
 * `floor/ceil(x)`
 * `len(x)` length of x
 * `sort(x)`
 * `pow(a, b)`
 * `parse(json)`
 * `stringify(json)`
 * `round(x)`
 * `sum(x)` compute sum of values in x
 * `codepoint(x)` get integer value of character `x`, where x is a string of length 1

I'm not sure if we system put them under a namespace like `builtin.pow` or `math.pow` or if we should name the otherwise, or which functions we might be missing...

But we probably need things like getting the length of a list, converting integer to string. Those we can't live without. I suspect a few others like uppercase/lowercase are useful too.

jonasfj:
Note:
 * I have no idea how many functions we want built-in
 * Which functions is too much
 * Which functions we might be missing

Another option is to say that strings have some properties like: `"hello".length()`, `"hello".uppercase()`
`"hello".lowercase()` and similarly, numbers have some properties like `4.toString()` and arrays have `[1,2,3].length()`... That might be better as those names won't ever conflict with variable names in the `context`.

We could probably define things like:
```js
let builtinMethods = {}

builtinMethods.string = {}
builtinMethods.string.length = function() {
  return this.length;
}
```

Then inside our interpreter when we parse `expression '.' identifier` we do don't throw an error if the type isn't object... then we lookup in builtinMethods first, and if nothing is found there we throw an error.

hammad13060:
I like the idea of having builtin helper methods 👍 
I'll rather go with the first approach because It is highly unlikely that someone will chose ```toLower``` or ```toUpper``` as property name. Along with that we will mention the built-in methods in the docs so that users of the library will choose the names accordingly. We can have methods in separate namespace if you want, lets say namespace be ```builtin``` so users only have to take care that none of the properties conflict with ```builtin```.

jonasfj:
it depends how many built-in methods we add...
Also I suspect that people who uses json-e will inject some custom built-in methods into the context...

[...]

PR #37, opened 2017-02-22
Standard library as base of context

hammad13060:
PR to add support for builtin functions in context. Inspired from discussion #30 .


src/builtins.js
```diff
@@ -0,0 +1,50 @@
...
+let builtins = {};
+
+// attaching math functions
+let mathBuiltins = ['min', 'max', 'sqrt', 'ceil', 'floor', 'abs'];
+
+mathBuiltins.forEach(prop => {
+  if (Math[prop] == undefined) {
+    throw new Error(`${prop} in Math undefined`);
+  }
+  builtins[prop] = Math[prop];
```
jonasfj: Let's assert input is numbers...
Also assert number of arguments...

src/builtins.js
```diff
@@ -0,0 +1,50 @@
+import {
+  isString, isNumber, isBool,
+  isArray, isObject, isFunction,
+} from './type-utils';
+import ExtendableError from 'es6-error';
+
+class BuiltinError extends ExtendableError {
```
jonasfj: Let's make an `errors.js` file with all the error classes...

src/builtins.js
```diff
@@ -0,0 +1,50 @@
...
+let mathBuiltins = ['min', 'max', 'sqrt', 'ceil', 'floor', 'abs'];
+
+mathBuiltins.forEach(prop => {
+  if (Math[prop] == undefined) {
+    throw new Error(`${prop} in Math undefined`);
+  }
+  builtins[prop] = Math[prop];
+});
+
+builtins['lowercase'] = str => {
```
jonasfj: What if we do:
```js
let builtins = {};
let define = (name, {arguments = [], variadic = null, invoke}) => (...args) => {
  if (!variadic && args.length > arguments.length) {
    // Error too many arguments...
  }
  // More checks for types etc...
  return invoke(args...);
};

define('lowercase', {
  arguments: ['string'],
  invoke: (s) => s.toLowerCase();
});

define('min', {
  variadic: 'number',
  invoke: (...args) => Math.min(args...);
});

export default builtins;
```
Then we can put all the type checks and checks for number of arguments in one place... and just declare the signature of the functions...

specification.yml
```diff
@@ -520,3 +520,385 @@ context:  {}
...
+error: true
+---
+title: TypeError uppercase (4)
+context: {}
+template: {$eval: 'uppercase([])'}
+error: true
+---
+title: TypeError uppercase (5)
+context: {key: 1}
+template: {$eval: 'uppercase(key)'}
```
jonasfj: We need a few test cases that methods can be overwritten...
example:
```
title: TypeError uppercase can be overwritten
context: {uppercase: 6}
template: {$eval: 'uppercase + 5'}
result: 11
```

src/index.js
```diff
@@ -176,6 +177,7 @@ let render = (template, context) => {
 
 module.exports = (template, context = {}) => {
   let test = Object.keys(context).every(v => /[a-zA-Z_][a-zA-Z0-9_]*/.exec(v)[0]);
+  Object.assign(context, builtins);
```
jonasfj: You shouldn't modify the context object... and builtins shouldn't overwrite existing context variables...
`context = Object.assign({}, builtins, context);`

src/builtins.js
jonasfj: I think we should use `define` proposed below...

src/builtins.js
jonasfj: Note it's possible that you want to use different options than `arguments` and `variadic`... Just was just a quick proposal, I think we can find smarter names and possibly more options to use for making this flexible...

hammad13060: COOL!  Idea looks really neat. I think we can do something like this as well. what do you say
```js
let builtins = {};
let define = (name, context, {arguments = [], variadic = null, invoke}) => context[name] = (...args) => {
  if (!variadic && args.length > arguments.length) {
    // Error too many arguments...
  }
  // More checks for types etc...
  return invoke(args...);
};

define('lowercase', builtins, {
  arguments: ['string'],
  invoke: (s) => s.toLowerCase();
});

define('min', builtins, {
  variadic: 'number',
  invoke: (...args) => Math.min(args...);
});

export default builtins;
```

hammad13060:
ping @jonasfj

src/builtins.js
jonasfj: FYI, no need to give `context` as parameter... You can just reference it as a global variable form inside the define function.... After all define isn't exported and only used to define builtins in this file...

src/builtins.js
```diff
@@ -0,0 +1,95 @@
...
+
+let builtins = {};
+
+let define = (name, context, {argumentTypes = [], variadic = null, invoke}) => context[name] = (...args) => {
+  if (!variadic && args.length < argumentTypes.length) {
+    throw builtinError(`builtin: ${name}`, `${args.toString()}, arguments too less`);
+  }
+
+  if (variadic) {
+    argumentTypes = args.map(() => variadic);
```
jonasfj: Neat trick to not have to consider varidaric...

src/builtins.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  let getArgumentTests = types => types.split('|').map(type => {
```
jonasfj: How about globally defining a map from typeName to function that checks it...
Also you are missing types: 'json' which is probably useful for something :)

Or even better, instead of specifying `argumentTypes: ` as strings how about: `arguments: [isString]`... Hence we just specify functions that return true/false... Then we can do any type check we want... We just import those type check functions from utils...

src/builtins.js
```diff
@@ -0,0 +1,95 @@
...
+      case 'number':   return isNumber;
+      case 'array':    return isArray;
+      case 'object':   return isObject;
+      case 'function': return isFunction;
+      case 'boolean':  return isBool;
+      default:         throw new Error(`${type} not supported for builtin arguments`);
+    }
+  });
+
+  args.forEach((arg, i) => {
```
jonasfj: This should just be: `if(!arguments[i](arg)) { throw....` then...

src/builtins.js
```diff
@@ -0,0 +1,95 @@
...
+      throw builtinError(`builtin: ${name}`, `argument ${i + 1} to be ${argumentTypes[i]} found ${typeof arg}`);
+    }
+  });
+
+  return invoke(...args);
+};
+
+// attaching math functions
+let mathBuiltins = [
+  ['min', builtins, {
```
jonasfj: Or just do:
`['min', 'max'].forEach(name => define(name, {....` and the same for the four others... That's much shorter... :)

src/builtins.js
```diff
@@ -0,0 +1,95 @@
...
+  invoke: str => str.toLowerCase(),
+});
+
+define('uppercase', builtins, {
+  argumentTypes: ['string'],
+  invoke: str => str.toUpperCase(),
+});
+
+define('str', builtins, {
+  argumentTypes: ['string|number|boolean|array'],
```
jonasfj: So if we passed functions instead of strings we could just make a special function for this case...

src/builtins.js
jonasfj: I'm wrong we can do this... Because then we can't print error messages with the type...

But we can make a global map from typeName to function checking it... Then we don't need the | syntax...

src/builtins.js
hammad13060: can you elaborate on this one.

src/builtins.js
hammad13060: I didn't quite get this. Can you elaborate a bit.

src/builtins.js
hammad13060: Ahh, nice observation :)

src/builtins.js
hammad13060: I am thinking of this

```javascript
import type_utils from './type-utils';

let define = (name, context, {argumentTests = [], variadic = null, invoke}) => context[name] = (...args) => {
  if (!variadic && args.length < argumentTests.length) {
    throw builtinError(`builtin: ${name}`, `${args.toString()}, arguments too less`);
  }

  if (variadic) {
    argumentTests = args.map(() => variadic);
  }

  args.forEach((arg, i) => {
    if (!argumentTests[i].split('|').some(test => type_utils[test](arg))) {
      throw builtinError(`builtin: ${name}`, `argument ${i + 1} to be ${argumentTests[i]} found ${typeof arg}`);
    }
  });

  return invoke(...args);
};

define('len', builtins, {
  argumentTests: ['isString|isArray'],
  invoke: obj => obj.length,
});
```

src/builtins.js
jonasfj:
```js
let types = {
  'string|number|bool|array': (value) => true || false...
}

define('str', {
  arguments: ['string|number|bool|array'],
  ...
});
```

src/type-utils.js
```diff
@@ -1,8 +1,9 @@
...
-  isArray:    (expr) => expr instanceof Array,
-  isObject:   (expr) => expr instanceof Object,
-  isFunction: (expr) => expr instanceof Function, 
+  isString:   expr => typeof expr === 'string',
+  isNumber:   expr => typeof expr === 'number',
+  isBool:     expr => typeof expr === 'boolean',
+  isArray:    expr => expr instanceof Array,
+  isObject:   expr => expr instanceof Object,
+  isFunction: expr => expr instanceof Function,
+  isJSON:     expr => expr instanceof JSON,
```
jonasfj: that won't work :)

hammad13060: I think this should work
``` javascript
let utils = {
  isString:   expr => typeof expr === 'string',
  isNumber:   expr => typeof expr === 'number',
  isBool:     expr => typeof expr === 'boolean',
  isArray:    expr => expr instanceof Array,
  isObject:   expr => expr instanceof Object,
  isFunction: expr => expr instanceof Function,
  isJSON:     expr => {
    if (utils.isString(expr) || utils.isNumber(expr) || utils.isBool(expr)) {
      return true;
    }

    if (utils.isArray(expr)) {
      return expr.every(v => utils.isJSON(v));
    }

    let result = true;
    if (utils.isobject(expr)) {
      for (let key of Object.keys(expr)) {
        if (expr.hasOwnProperty(key)) {
          result = result && utils.isJSON(expr[key]);
          if (!result) {
            break;
          }
        }
      }
    }
    return result;
  },
};

module.exports = utils;

```

hammad13060:
For supporting multiple argument types I am thinking of going for this

```javascript
import type_utils from './type-utils';

let define = (name, context, {argumentTests = [], variadic = null, invoke}) => context[name] = (...args) => {
  if (!variadic && args.length < argumentTests.length) {
    throw builtinError(`builtin: ${name}`, `${args.toString()}, arguments too less`);
  }

  if (variadic) {
    argumentTests = args.map(() => variadic);
  }

  args.forEach((arg, i) => {
    if (!argumentTests[i].split('|').some(test => type_utils[test](arg))) {
      throw builtinError(`builtin: ${name}`, `argument ${i + 1} to be ${argumentTests[i]} found ${typeof arg}`);
    }
  });

  return invoke(...args);
};

define('len', builtins, {
  argumentTests: ['isString|isArray'],
  invoke: obj => obj.length,
});
```

src/builtins.js
```diff
@@ -0,0 +1,67 @@
...
+  invoke: str => str.toUpperCase(),
+});
+
+define('str', builtins, {
+  argumentTests: ['isString|isNumber|isBool|isArray'],
+  invoke: obj => obj.toString(),
+});
+
+define('len', builtins, {
+  argumentTests: ['isString|isArray'],
```
jonasfj: You'll probably want to change these names to `string|array` otherwise it won't read well in error messages..

how about we have mapping:
```js
let types = {
  string: isString,
  array: isArray,
  ...
};
```

That way we can write `'array|string'` and still have nice error messages...

jonasfj:
Otherwise I think we're almost good to land... Oh, maybe fix merge conflicts :)

hammad13060:
Done :)

Issue #38, opened 2017-02-22
Demo website for json-e


jonasfj:
We should make a github pages site that looks this:

![https://img42.com/0b78N](https://cdn.img42.com/74ea5a4dc0e90396733473cc23fa14a3.png)

With a live "try it" editor... that way we can easily test bugs etc...

hammad13060:
Seems like a good Idea. I don't have much idea about front end development, so can you give some pointers on how to start ? Do we have some ```jekyll``` templates which server our purpose ?

jonasfj:
I figure we can built it with neutrino or something... So we get everything
bundled into a single js file...

We'll probably have to do it a subfolder...As it'll need it's own
package.json... Hmm maybe not sure how we structure it.. maybe devDeps will
do...

Ping me on IRC... Eli is sure to have some ideas for how we can quickly
hack this together... IMO we want it to be super easy to update, so we keep
it up to date...

Den 22. feb. 2017 7.49 PM skrev "Hammad Akhtar" <notifications@github.com>:

jonasfj:
@hammad13060 Hi, sorry I missed on you on IRC.
Thinking about we can probably just use http://browserify.org/ to compile all the files into a single bundle that we can load into an HTML... then we don't need anything complicated.

You add `browserify` and `codemirror`, `jquery`, etc... as `devDependencies` in `package.json`. 
Then create files `demo.js` and `index.html`, in `demo.js` we just write a script that compile with browserify bundles up everything and then we include that `index.html`. We can use `npm run build-demo` to build browserify thing, we'll just add a script command in `package.json`.

hammad13060:
Hmm..this looks good :)

Just few clarifications
1. first we will build josn-e using ```npm run compile```
2. then we will include ```lib/index.js``` to out ```demo.js```
3. then we will run ```npm run build-demo```
4. now we are good to go
am I right ?

What directory structure should we follow for website ?

hammad13060:
Hi @jonasfj I ran into following issue 

I did following steps
1. ```npm install browserify -g```
2. ```npm install codemirror --save-dev```

Now my ```demo.js``` looks like this

```javascript
let CodeMirror = require('codemirror');
let editor = CodeMirror(document.getElementById("try-it-template"), {
  value: 'let a = 123\n',
  mode: 'javascript',
});
```
Finally I transpile ```demo.js``` to ```budnle.js``` using ```browserify demo.js -o bundle.js```

[...]

PR #43, opened 2017-05-08
Adding json-e website

TheNavigat:
Apparently, this is a very basic implementation of a demo website for json-e.

I'm not sure if that's good enough to be merged (with later improvements) or if there are specific things we should fix before we merge. Feel free to comment on any such issues.

A live sample of the site is available at https://thenavigat.github.io/json-e/.

### DIFF
--- a/docs/index.js
+++ b/docs/index.js
@@ -1,0 +1,28 @@
+"use strict";
+
+var _lib = require("../lib");
+
+var _lib2 = _interopRequireDefault(_lib);
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+$(function() {
+    //hang on event of form with id=myform
+    $("form").submit(function(e) {
+        //prevent Default functionality
+        e.preventDefault();
+
+        var template = $("#template").val();
+        var context = $("#context").val();
+
+        eval("template" + "=" + template);
+        eval("context" + "=" + context);
+
+        var output = (0, _lib2.default)(template, context);
+
+        $("textarea").val(JSON.stringify(output));
+
+        return false;
+    });
+
+});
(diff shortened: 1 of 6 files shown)


docs/index.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+var _lib = require("../lib");
+
+var _lib2 = _interopRequireDefault(_lib);
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```
djmitche: It looks like tihs file has been transpiled with babel?  Since this is a source file (as opposed to bundle.js), it should be OK to write it in native ES6 and let babel-compile transpile it.

docs/index.js
```diff
@@ -0,0 +1,28 @@
...
+    //hang on event of form with id=myform
+    $("form").submit(function(e) {
+        //prevent Default functionality
+        e.preventDefault();
+
+        var template = $("#template").val();
+        var context = $("#context").val();
+
+        eval("template" + "=" + template);
+        eval("context" + "=" + context);
```
djmitche: I'd rather not use `eval` for parsing JSON-like data structures.

Let's use YAML, instead?

jonasfj:
Have you considered embedding codemirror?
And using YAML as input language....

Also provide a few examples.

Note: Just use bootstrap or something like it to style your buttons.

jonasfj:
Note: it doesn't have to follow the mockup proposed in https://github.com/taskcluster/json-e/issues/38. But maybe take a look at it... I think that'll provide much more educational value for people playing with it.

jonasfj:
The goal of this tool is:
 * To be educational for people learning json-e, by demonstrating features
 * To be useful for people using json-e to write templates by allowing us to test snippets quickly.

docs/index.js
TheNavigat: Done. Thanks for that comment!

djmitche:
OK, I broke the rest of the bits out into new issues.

PR #49, opened 2017-05-09
document the language, so far, in the README

djmitche:
@jonasfj, does this match your understanding?

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,8 @@ unusual case to include a JSON string in a larger data structure.
 
 ```yaml
 context:  {a: 1, b: 2}
-template: {$json: [a, b, c, 1, 2, 3]}
-result:   '["a", "b", "c", 1, 2, 3]'
+template: {$json: [a, b, {$eval: 'a+b', 4]}
+result:   '["a", "b", 3, 4]'
 ```
 
 ### `$if` - `then` - `else`
@@ -118,6 +118,24 @@ omitted, then the expression is omitted from the parent object.
 context:  {cond: true}
 template: {key: {$if: 'cond', then: 1}, k2: 3}
 result:   {key: 1, k2: 3}
+```
+
+```yaml
+context:  {x: 10}
+template: {$if: 'x > 5', then: 1, else: -1}
+result:   1
+```
+
+```yaml
+context: {cond: false}
+template: [1, {$if: 'cond', else: 2}, 3]
+result: [1,2,3]
+```
+
+```yaml
+context: {cond: false}
+template: {key: {$if: 'cond', then: 2}, other: 3}
+result: {other: 3}
 ```
 
 ### `$fromNow`
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ unusual case to include a JSON string in a larger data structure.
 
 ```yaml
 context:  {a: 1, b: 2}
-template: {$json: [a, b, {$eval: 'a+b', 4]}
+template: {$json: [a, b, {$eval: 'a+b'}, 4]}
 result:   '["a", "b", 3, 4]'
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,8 +31,188 @@ console.log(jsone(template, context));
...
+
+### `$json`
+
+The `$json` operator formats the given value as JSON. It does not evaluate the
+value (use `$eval` for that). While this can be useful in some cases, it is an
+unusual case to include a JSON string in a larger data structure.
+
+```yaml
+context:  {a: 1, b: 2}
+template: {$json: [a, b, c, 1, 2, 3]}
```
jonasfj: That would be the case without `$json` too... the case you want to show is:
`{$json: [1,2, {$eval: 'a+b'}]}` => `[1,2, {$eval: 'a+b'}]`, while `[1,2, {$eval: 'a+b'}]` => `[1,2,3]`

README.md
```diff
@@ -25,8 +31,188 @@ console.log(jsone(template, context));
...
+
+### `$if` - `then` - `else`
+
+The `$if` operator supports conditionals. It evalates the given value, and
+replaces itself with the `then` or `else` properties. If either property is
+omitted, then the expression is omitted from the parent object.
+
+```yaml
+context:  {cond: true}
+template: {key: {$if: 'cond', then: 1}, k2: 3}
```
jonasfj: This miss one important case: `else`

It also miss the use-case with lists and objects where an omitted `then` or `else` case becomes a delete marker.
```yaml
context: {cond: false}
template: [1, {$if: 'cond', then: 2}, 3]
result: [1,3]
```
and
```yaml
context: {cond: false}
template: [1, {$if: 'cond', else: 2}, 3]
result: [1,2,3]
```

```yaml
context: {cond: false}
template: {key: {$if: 'cond', then: 2}, other: 3}
result: {other: 3} // without the 'else' branch {$if: 'cond', then: 2} becomes a delete-marker and 'key' disappears
```

README.md
```diff
@@ -25,8 +31,206 @@ console.log(jsone(template, context));
...
+
+### `$json`
+
+The `$json` operator formats the given value as JSON. It does not evaluate the
+value (use `$eval` for that). While this can be useful in some cases, it is an
+unusual case to include a JSON string in a larger data structure.
+
+```yaml
+context:  {a: 1, b: 2}
+template: {$json: [a, b, {$eval: 'a+b', 4]}
```
TheNavigat: I think there's a missing `}` here

README.md
```diff
@@ -25,8 +31,206 @@ console.log(jsone(template, context));
...
+* Arithmetic on integers (`+`, `-`, `*`, `/`, `**` for exponentiation), with unary `-` and `+`
+* String concatenation (`+`)
+* Comparison of strings to strings or numbers to numbers (`<`, `<=`, `>`, `>=`)
+* Equality of anything (`==`, `!=`)
+* Boolean operators (`||`, `&&`, `!`)
+* Identifiers referring to variables (matching `/[a-zA-Z_][a-zA-Z_0-9]*/`)
+* Object property access: `obj.prop` or `obj["prop"]`
+* Array indexing and slicing
+  * `array[1]` -- second element of array (zero-indexed)
+  * `array[1:3]` -- second and third elements of the array
```
TheNavigat: That's a bit ambiguous. I'd prefer if it's rather the following: `array[1:4]` -- second through fourth element of the array

README.md
```diff
@@ -25,8 +31,206 @@ console.log(jsone(template, context));
...
+* Arrays in JSON format (`[.., ..]`)
+* Parentheses for grouping (`(a + b) * c`)
+* Arithmetic on integers (`+`, `-`, `*`, `/`, `**` for exponentiation), with unary `-` and `+`
+* String concatenation (`+`)
+* Comparison of strings to strings or numbers to numbers (`<`, `<=`, `>`, `>=`)
+* Equality of anything (`==`, `!=`)
+* Boolean operators (`||`, `&&`, `!`)
+* Identifiers referring to variables (matching `/[a-zA-Z_][a-zA-Z_0-9]*/`)
+* Object property access: `obj.prop` or `obj["prop"]`
+* Array indexing and slicing
```
TheNavigat: Based on my understanding, array[1:3] returns the second (array[1]) and third (array[2]) elements. It would be nice to add a comment that the range indexing is closed on the starting end and open on the last end. In simpler more mathematical words, it would be nice to add a comment that range selection is [  [.

djmitche:
Good points, thanks!

README.md
jonasfj: I suspect it's better to say it has Python semantics :)

PR #53, opened 2017-05-10
consolidate all tests into specification.yml

djmitche:
This puts all of the test cases into the same file.  I've done some renaming from "foo (17)" to a more descriptive name, but quickly got bored.  Sorry :)

I think it's important that all of these values be in specification.yml, so that the Python implementation is tested against the same cases as the JS implementation.


jonasfj:
There already is a concept of sections in the file...
IMO splitting it or not is probably nit picking...
Note: it easier to download into other repos when implementing in Python
etc...

Den 10. maj 2017 4.55 PM skrev "Amjad Mashaal" <notifications@github.com>:

djmitche:
> Before you read my comment, I'm sorry for almost always having a different opinion. I come from a different background, you know...

Don't apologize for that! 

I don't mind either way.  Jonas makes the point that it will be easier to have a single file when we implement this in Python, but only a little bit easier.  Let's do that Python implementation, and afterward if we decide it'd be nicer to break it up (maybe into `specification/json.yml`, `specification/eval.yml`, etc.?) we can do that for both languages at the same time.

TheNavigat:
@jonasfj is correct, it's easier to have a single file, but my argument is why can't we do both?

> IMHO, a smarter way to do it would be to split the file into smaller files (including the ones deleted), and adding a script that consolidates all the smaller test files into specification.yml. Of course, this requires more uniformity among the child test files.

Editing smaller files, each relevant to its own section, makes more sense to me than editing one huge file (even if it contains sections), and consolidation of these files into specification.yml makes it easier to require all that data in other projects (such as the Python implementation), but the consolidation can be done automatically.

This is analogous to what Certbot's team has done here, https://github.com/certbot/certbot/tree/master/letsencrypt-auto-source. The `letsencrypt-auto.template` is the template for the consolidated `letsencrypt-auto` bash file, where the pieces exist in the `pieces` directory. You can always download that single file for usage, but when you're editing it, you should edit the pieces and rebuild.

What are your thoughts?

djmitche:
OK, well, let's merge this change (I think this is important because `tests/expressions.yml` and `specification.yml` suggests that the former isn't part of the specification), and then you can make a PR to break it up into `specification/*.yml`.  I don't think we need to combine the files -- and especially not to combine them and check in the result.  Rather, then JS and Python implementations can just read from the multiple files directly.

PR #55, opened 2017-05-11
Don't upload the gh-pages docs/ directory to docs.tc.net

djmitche:
Instead, specify DOCS_FOLDER=tc-docs, which doesn't exist. Then we will
only upload the README.


djmitche:
Github can only look in docs/ - it doesn't let you specify.

Issue #59, opened 2017-05-13
Add $merge


TheNavigat:
This is part of https://github.com/taskcluster/json-e/issues/10.

I agree with @jonasfj 's statement:

> I'm thinking `{$merge: [{a: 1, b: 1}, {b: 2}]} = {a: 1, b: 2}`, ie. right side overwrites.
$mergeDeep is the same except it does it recursively to all objects.

PR #60, opened 2017-05-14
Adding `$merge`

TheNavigat:
Closes https://github.com/taskcluster/json-e/issues/59

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,9 +167,17 @@ template:
 result:   [3, 5, 7]
 ```
 
-The array is the value of the `$map` property, and the expression to evaluate
-is given by `each(var)` where `var` is the name of the variable containing each
-element.
+### `$merge`
+
+The `$merge` constructor merges an array of objects, returning a single object
+that combines all of the objects in the array, where the right-side objects
+overwrite the values of the left-side ones.
+
+```yaml
+context:  {}
+template: {$merge: [{a: 1, b: 1}, {b: 2, c: 3}, {d: 4}]}
+result:   {a: 1, b: 2, c: 3, d: 4}
+```
 
 ### `$sort`
 
--- a/specification.yml
+++ b/specification.yml
@@ -55,6 +55,11 @@ context:  {}
 template: {$json: [1,2,true,{},[]]}
 result:   '[1,2,true,{},[]]'
 ---
+title:    $merge
+context:  {}
+template: {$merge: [{a: 1, b: 1}, {b: 2, c: 3}, {d: 4}]}
+result:   {a: 1, b: 2, c: 3, d: 4}
+---
 title:    string interpolation
 context:  {key: 'world', num: 1}
 template: {message: 'hello ${key}', 'k-${num}': true}
--- a/src/index.js
+++ b/src/index.js
@@ -84,6 +84,26 @@ constructs.$map = (template, context) => {
 
 };
 
+constructs.$merge = (template, context) => {
+  let value = render(template['$merge'], context);
+
+  if (!isArray(value) && !isArray(template['$merge'])) {
+    throw jsonTemplateError('$merge value must evaluate to an array\n', template);
+  }
+
+  if (!isArray(value)) {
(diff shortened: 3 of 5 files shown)


src/index.js
```diff
@@ -105,6 +105,26 @@ constructs.$map = (template, context) => {
 
 };
 
+constructs.$merge = (template, context) => {
+  let value = render(template['$merge'], context);
+
+  if (!isArray(value) && !isArray(template['$merge'])) {
```
jonasfj: Explain why you have `!isArray(template['$merge'])`

src/index.js
```diff
@@ -105,6 +105,26 @@ constructs.$map = (template, context) => {
...
+
+  if (!isArray(value)) {
+    throw jsonTemplateError('$merge requires array as value\n', template);
+  }
+
+  let obj = {};
+
+  value.forEach(function(x) {
+    Object.assign(obj, x);
+  });
```
jonasfj: Can't you do: `return Object.assgin({}, ...value);` or `return Object.assgin({}, ...value.reverse());` (you probably need to check semantics of `Object.assign`

src/index.js
jonasfj: IMO we only care what it evaluates to... not how it got to be that value...

TheNavigat: You're correct, I deleted it that check.

src/index.js
TheNavigat: I'm not sure how to do that. I checked `Object.assign`'s documentation and couldn't find any information on how to assign an array of objects in a single line.

When I tried, this happened.

```javascript
> x = [{a: 1}, {a: 2}, {a: 3}]
[ { a: 1 }, { a: 2 }, { a: 3 } ]
> Object.assign({}, x);
{ '0': { a: 1 }, '1': { a: 2 }, '2': { a: 3 } }
```

jonasfj:
`Object.assign({}, a, b, ...c)` is valid js... So is where c is a list of
objects...

Den 21. maj 2017 7.05 PM skrev "Amjad Mashaal" <notifications@github.com>:

jonasfj:
As you can see here sources can be more than one:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

Den 21. maj 2017 8.49 PM skrev "Jonas Finnemann Jensen" <jopsen@gmail.com>:

TheNavigat:
@jonasfj If c is a list that doesn't work, nope.

`Object.assign({}, a, b, c);` works but I'm not sure how can I execute the command with `a, b, c` where those arguments are actually a list. Can you show me how to do that, please?

jonasfj:
`Object.assign({}, ...[a, b, c]);`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator

TheNavigat:
I never knew that existed. I assumed the `...` you typed above meant the `...` used in English, not as an operator. I apologize :)

Code is updated. Please let me know if there's anything I should fix/improve.

README.md
```diff
@@ -167,9 +167,17 @@ template:
 result:   [3, 5, 7]
 ```
 
-The array is the value of the `$map` property, and the expression to evaluate
-is given by `each(var)` where `var` is the name of the variable containing each
-element.
```
djmitche: This paragraph of documentation for `$map` appears to have disappeared..

specification.yml
```diff
@@ -55,6 +55,11 @@ context:  {}
 template: {$json: [1,2,true,{},[]]}
 result:   '[1,2,true,{},[]]'
 ---
+title:    $merge
+context:  {}
+template: {$merge: [{a: 1, b: 1}, {b: 2, c: 3}, {d: 4}]}
+result:   {a: 1, b: 2, c: 3, d: 4}
+---
```
djmitche: We should probably test some edge cases, too.  What does `{$merge: []}` return?  How about `{$merge: [{}]}`?

PR #70, opened 2017-05-16
Sort numbers numerically

djmitche:
Numbers should sort in numeric order -- they are currently sorted lexically.

Strings should sort in lexical order -- this works now.

Sorting between the two, without a `by` function, can probably remain undefined (and we don't need to check for that case).


djmitche:
Whoops, igore this.

PR #44, opened 2017-05-16
Add tests for some corner cases

djmitche:
I'll be making a list here as I think of them..

 * [x] division applies floor (or array indexes don't require ints)
 * [x] string interpolation results in a boolean
 * [x] string interpolation results in null
 * [x] string interpolation escaping
 * [x] operator escaping (how do I write a literal `{$eval: foo}`)?
 * [x] object access by named property: `{$eval: "settings[environment]"}`
 * [x] for `$if`, what values are truthy/falsy (empty strings, 0, [], {}, null)
 * [x] access of missing property in expression (`"x.foo"` when `x` has no `foo`)
 * [x] missing `then` in `$if`-`else` expression
 * [x] top-level deletion marker (`{$if: 'false', then: 1}`)
 * [x] deletion marker in an array (`[1, 2, {$if: 'false', then: 1}]`)
 * [x] `$map` on non-array things other than an object
 * [x] `$sort` over multi-digit numbers (1 < 10)
 * [x] `$sort` of mixed numbers and strings and other objects
 * [x] `$sort` of mixed numbers and strings
 * [x] `$sort` of all non-array types
 * [x] `$sort` when `by(var)` returns various types
 * [x] `$reverse` when given a non-array
 * [x] expressions: `==` does deep equality, `!=` does the opposite
 * [x] `$eval` given things other than array, string
 * [x] Numeric literals - floats, different bases, leading zeroes
 * [x] parenthetical grouping
 * [x] all parser operators (infix and unary) over all types
 * [x] string literals with `"` and with `'`
 * [x] array indexing / slicing with invalid types
 * [x] object property access with invalid types
 * [x] min/max with zero args
 * [x] all builtins with funny types
 * [x] `str([..])` formatting
 * [x] boolean literals (true, false)

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -30,16 +30,6 @@ context:  {}
 template: {$fromNow: '2 days 3 hours'}
 result:   '2017-01-21T19:27:20.974Z'
 ---
-title:    fromNow built-in
-context:  {}
-template: {$eval: fromNow("")}
-result:   '2017-01-19T16:27:20.974Z'
----
-title:    fromNow built-in 2 days 3 hours
-context:  {}
-template: {$eval: fromNow("2 days 3 hours")}
-result:   '2017-01-21T19:27:20.974Z'
----
 title:    $eval
 context:  {key: ['value', 1, 2, true, {}]}
 template: {$eval: 'key'}
@@ -693,6 +683,11 @@ context: {key1: 1, key2: 2}
 template: {$eval: 'min(2, 1, 3, 4, 5)'}
 result: 1
 ---
+title: min - TypeError
+context: {}
+template: {$eval: 'min("11")'}
+error: true
+---
 title: min (7)
 context: {}
 template: {$eval: 'min()'}
@@ -729,6 +724,11 @@ context: {key1: 1, key2: 2}
 template: {$eval: 'max(2, 1, 3, 4, 5)'}
 result: 5
 ---
+title: max - TypeError
+context: {}
+template: {$eval: 'max("12")'}
+error: true
+---
 title: max (7)
 context: {}
 template: {$eval: 'max()'}
@@ -754,6 +754,11 @@ title: sqrt (3)
 context: {}
 template: {$eval: 'sqrt(9)'}
 result: 3.0
+---
+title: sqrt - TypeError
+context: {}
+template: {$eval: 'sqrt("nine")'}
+error: true
 ---
 title: ceil (1)
 context: {key: 1.1}
@@ -795,6 +800,11 @@ context: {key: 1.0}
 template: {$eval: 'ceil(key)'}
 result: 1.0
 ---
+title: ceil - TypeError
+context: {}
+template: {$eval: 'ceil(true)'}
+error: true
+---
 title: floor (1)
 context: {key: 1.1}
 template: {$eval: 'floor(key)'}
@@ -834,6 +844,11 @@ title: floor (8)
 context: {key: 1.0}
 template: {$eval: 'floor(key)'}
 result: 1.0
+---
+title: floor - TypeError
+context: {key: null}
+template: {$eval: 'floor(key)'}
+error: true
 ---
 title: abs (1)
 context: {key: -1.1}
@@ -934,6 +949,11 @@ title: abs (20)
 context: {}
 template: {$eval: 'abs(1.0)'}
 result: 1.0
+---
+title: abs - TypeError
+context: {}
+template: {$eval: 'abs({})'}
+error: true
 ---
 title: lowercase (1)
 context: {key: 'HEllo'}
@@ -1095,6 +1115,11 @@ context: {}
 template: {$eval: 'str(true)'}
 result: 'true'
 ---
+title: str (6)
+context: {x: null}
+template: {$eval: 'str(x)'}
+error: true
+---
 title: str (7)
 context: {key: false}
 template: {$eval: 'str(key)'}
@@ -1114,6 +1139,21 @@ title: str (10)
 context: {}
 template: {$eval: 'str([1, 2, 3])'}
 result: '1,2,3'
+---
+title:    fromNow
+context:  {}
+template: {$eval: fromNow("")}
+result:   '2017-01-19T16:27:20.974Z'
+---
+title:    fromNow - 2 days 3 hours
+context:  {}
+template: {$eval: fromNow("2 days 3 hours")}
+result:   '2017-01-21T19:27:20.974Z'
+---
+title:    fromNow - TypeError
+context:  {}
+template: {$eval: fromNow(13)}
+error: true
 ---
 title: override builtin (1), uppercase
 context: {uppercase: 1}
--- a/specification.yml
+++ b/specification.yml
@@ -80,6 +80,17 @@ context:  {a: 'hello', b: 'world'}
 template: {message: '${a}##${b}'}
 result:   {message: 'hello##world'}
 ---
+title: string interpolation escapes
+context:  {}
+template: {message: 'a literal \\${ in a string'}
+result:   {message: 'a literal ${ in a string'}
+todo: 'https://github.com/taskcluster/json-e/issues/66'
+---
+title: string interpolation with unbalanced }
+context:  {}
+template: {message: 'tricky ${"}}}}"}'}
+result:   {message: 'tricky }}}}'}
+---
 title:    can't interpolate arrays
 context:  {key: [1,2,3]}
 template: {message: 'hello ${key}'}
--- a/specification.yml
+++ b/specification.yml
@@ -602,6 +602,31 @@ context:  {}
 template: {$reverse: [3, 4, 1, 2]}
 result:   [2, 1, 4, 3]
 ---
+title:    reverse of an object
+context:  {}
+template: {$reverse: {a: 1}}
+error: true
+---
+title:    reverse of null
+context:  {}
+template: {$reverse: null}
+error: true
+---
+title:    reverse of a string
+context:  {}
+template: {$reverse: 'foo'}
+error: true
+---
+title:    reverse of a number
+context:  {}
+template: {$reverse: 1}
+error: true
+---
+title:    reverse of a boolean
+context:  {}
+template: {$reverse: true}
+error: true
+---
 title:    simple reverse with $eval
 context:  {key: [3, 4, 1, 2]}
 template: {$reverse: {$eval: 'key'}}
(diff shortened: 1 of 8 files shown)


djmitche:
OK, this is done:

```
  763 passing (828ms)
  37 pending
```

yeah, that's a lot, and yeah, many of them are super-boring, but they do test a lot of corner cases which might otherwise be different between different implementations of JSON-e.

Where I've found problems, I've added a `todo` pointing to the github issue to fix it.  So we can remove those "todo" items in the process of fixing the issue.

jonasfj:
More than 2k additions... I'll read that Monday :)

djmitche:
It's probably easiest to review commit-by-commit.

TheNavigat:
abided shows a read underline in Chrome. Abode is correct, https://www.merriam-webster.com/dictionary/abide. Apparently abided is also correct anyway :)

specification.yml
```diff
@@ -1371,6 +1371,21 @@ context: {key: {a: 1}}
...
+---
+title: 'missing property'
+context: {key: {a: 1}}
+template: {$eval: 'key.b'}
+error: true
+---
+title: 'missing property by name'
+context: {key: {a: 1}}
+template: {$eval: 'key["b"]'}
+error: true
```
jonasfj: I'm not sure about this one... returning `null` or `deleteMarker` might be more powerful

specification.yml
```diff
@@ -269,6 +269,12 @@ title: $if->then->else, nonempty object
 context: {cond: {a: 2}, key: 'world'}
 template: {$if: 'cond', then: "t", else: "f"}
 result: t
+---
+title: $if->then evaluating to nothing at the top level
+context: {cond: false}
+template: {$if: 'cond', then: "t"}
+result: {}
```
jonasfj: I advocate for `null`... that way `deleteMarker` is just null, but with some slightly modified semantics in list and objects...

djmitche: JSON distinguishes `{"x": null}` from `{}`, so I think we should too.

specification.yml
djmitche: That's a behavior that enables typos and surprises when data formats change. I prefer missing properties or keys to be an error.

specification.yml
```diff
@@ -2327,6 +2327,391 @@ title: 'complex logical operation (4)'
...
+error: true
+---
+title: 'number and number'
+context: {a: 345, b: 123}
+template: {$eval: "a && b"}
+error: true
+---
+title: 'number or number'
+context: {a: 345, b: 123}
+template: {$eval: "a || b"}
```
jonasfj: All of this is inconsistent with the fact that all other values could be used as true in `$if`

djmitche: It's true, and I just updated #68 to mention that.

specification.yml
```diff
@@ -422,6 +594,31 @@ context:  {}
...
+context:  {}
+template: {$reverse: {a: 1}}
+error: true
+---
+title:    reverse of null
+context:  {}
+template: {$reverse: null}
+error: true
+---
+title:    reverse of a string
```
TheNavigat: Why shouldn't we be able to reverse a string?

specification.yml
```diff
@@ -515,6 +762,11 @@ context: {}
 template: {$eval: 'sqrt(9)'}
 result: 3.0
 ---
+title: sqrt (TypeError)
```
TheNavigat: I think we need to add a test for `sqrt()`

specification.yml
```diff
@@ -555,6 +807,11 @@ context: {key: 1.0}
 template: {$eval: 'ceil(key)'}
 result: 1.0
 ---
+title: ceil (TypeError)
```
TheNavigat: `ceil()`

specification.yml
```diff
@@ -595,6 +852,11 @@ context: {key: 1.0}
 template: {$eval: 'floor(key)'}
 result: 1.0
 ---
+title: floor (TypeError)
```
TheNavigat: `floor()`

specification.yml
```diff
@@ -695,6 +957,11 @@ context: {}
 template: {$eval: 'abs(1.0)'}
 result: 1.0
 ---
+title: abs (TypeError)
```
TheNavigat: `abs()`

specification.yml
```diff
@@ -855,6 +1122,11 @@ context: {}
 template: {$eval: 'str(true)'}
 result: 'true'
 ---
+title: str (6)
```
TheNavigat: `str()`

specification.yml
```diff
@@ -1056,114 +1379,1411 @@ context: {a: 2, b: 3}
...
+title: 'multiplication of strings to strings'
+context: {a: 'a', b: 'b'}
+template: {$eval: 'a * b'}
+error: true
 ---
-title: 'and operator (1)'
-context: {tt: true, ff: false}
-template: {$eval: 'tt && tt'}
-result: true
+title: 'multiplication of strings to numbers'
```
TheNavigat: In a certain language (I can't recall which) that basically printed the string `b` times. Do you think that might be a better approach?

specification.yml
```diff
@@ -1056,114 +1379,1411 @@ context: {a: 2, b: 3}
...
+title: 'multiplication of strings to objects'
+context: {a: 'a', b: {x: 1}}
+template: {$eval: 'a * b'}
+error: true
 ---
-title: 'and operator with not operator'
-context: {tt: true, ff: false}
-template: {$eval: 'tt && !ff'}
-result: true
+title: 'multiplication of strings to arrays'
```
TheNavigat: Either change this to division, or change the `/` to `*`

specification.yml
```diff
@@ -1056,114 +1379,1411 @@ context: {a: 2, b: 3}
...
+context: {tt: true, ff: false}
+template: {$eval: '!ff'}
+result: true
+---
+title: 'or operator (1)'
+context: {tt: true, ff: false}
+template: {$eval: 'tt || ff'}
+result: true
+---
+title: 'or operator (2)'
```
TheNavigat: Redundant test, identical to `or operator (1)`

specification.yml
TheNavigat: Is it possible to add an operator or so to allow printing `null` if the property doesn't exist? In some cases that would be very useful, especially since I don't recall that we a have a "does property exist" functionality. Mmmmm, should we add that? :)

specification.yml
```diff
@@ -1724,6 +3481,413 @@ title: 'greater than equal (4)'
...
+---
+title: equality of object and boolean
+context: {a: {}, b: true}
+template: {$eval: 'a == b'}
+result: false
+---
+title: equality of object and object
+context: {a: {}, b: {}}
+template: {$eval: 'a == b'}
+result: false
```
TheNavigat: Why is this false?

specification.yml
```diff
@@ -1724,6 +3481,413 @@ title: 'greater than equal (4)'
...
+---
+title: equality of array and boolean
+context: {a: [], b: true}
+template: {$eval: 'a == b'}
+result: false
+---
+title: equality of array and array
+context: {a: [], b: {}}
```
TheNavigat: `array and array` should be `array and object`

specification.yml
```diff
@@ -1724,6 +3481,413 @@ title: 'greater than equal (4)'
...
+template: {$eval: 'a == b'}
+result: false
+---
+title: equality of array and array
+context: {a: [], b: []}
+template: {$eval: 'a == b'}
+result: false
```
TheNavigat: Why is this false?

specification.yml
jonasfj: but when `else` branch is missing in `{$if: 'false', then: "t"}`, wouldn't it be more correct to resolve as `null`.
Current semantics imply that: `{$if: 'false', then: "t"}` is the same as `{$if: 'false', then: "t", else: {}}` if used as top-level, but not if used inside an array...

`result: {}` is an accident... I'm pretty sure the `{}` given here is the object we call ´_deleteMarker` in the code.. .We use to signal delete-marker. In this case there is no `else` branch, so the property or array index that the value is assigned to should be deleted. But since this is top-level that won't happy.

Hence, my argument that if `_deleteMarker` is the top-level result, the result to should be transformed to `null`.

Hence, `{$if: 'false', then: "t"}` is the same as `{$if: 'false', then: "t", else: null}` if used as top-level object, both become `null`.

If instead we have `[{$if: 'false', then: "t"}]` and `[{$if: 'false', then: "t", else: null}]`, then the results are `[]` and `[null]` respectively.


-----
Current behavior:
 * `{$if: 'false', then: "t"} => {}` // happens by accident because `_deleteMarker == {}`
 * `{$if: 'false', then: "t", else: {}} => {}`
 * `{$if: 'false', then: "t", else: null} => null`
 * `[{$if: 'false', then: "t"}] => []` // intended behavior!!!
 * `[{$if: 'false', then: "t", else: {}}] => [{}]`
 * `[{$if: 'false', then: "t", else: null}] => [null]`
I suggest:
 * `{$if: 'false', then: "t"} => null` **changed this special case**
 * `{$if: 'false', then: "t", else: {}} => {}`
 * `{$if: 'false', then: "t", else: null} => null`
 * `[{$if: 'false', then: "t"}] => []`
 * `[{$if: 'false', then: "t", else: {}}] => [{}]`
 * `[{$if: 'false', then: "t", else: null}] => [null]`

specification.yml
jonasfj: make a function for it... the `$<keyword>` constructs are intended to operate on JSON structures... They aren't intended to be auxiliary functions...

specification.yml
jonasfj: propose the test in a PR.

specification.yml
jonasfj: good question, didn't we have deep equality?

specification.yml
jonasfj: > That's a behavior that enables typos and surprises when data formats change. I prefer missing properties or keys to be an error.

Any thoughts for a syntax for taking a property or falling back to a default...
The reason I like null is that often you'll want to do `key['property'] || 'default-value'`

djmitche: Hm, it's true, since we don't have conditional expressions, we have no way to express that (note that `||` currently requires booleans).

So I can see a few options:
 * `x['foo']` and `x.foo` are `null` - but this will not "go away" when users want it to
 * `x['foo']` and `x.foo` are a delete marker - this will require handling delete markers within the expression language, where currently they're only handled in array/object construction
 * `x['foo']` and `x.foo` are errors, but `get(x, 'foo', 123)` is `123`, and the `get` built-in always requires that third argument.
Preferences?

specification.yml
djmitche: Ah, nice catch!

I think I intended this case to be an error -- it will be unusual to use an `$if` at the top level, and a read of a one-armed, top-level `$if` leaves the other arm undefined.  If you have a reason for that not to be an error, I don't mind null, but I'd prefer an error.  The other cases in your list above I agree with.

specification.yml
djmitche: I can add one here :)

specification.yml
djmitche: I don't understand this suggestion - none of the other descriptions include the empty parentheses.  Why does it matter?

specification.yml
djmitche: Python does that.  i'm not sure it'd be useful, though.  We could add it later if it is.

specification.yml
djmitche: Oops, yes, that's definitely a mistake.

specification.yml
djmitche: Anyway, we can sort that out after landing this patch, and adjust the tests when we make the change.  We can do whatever we want until 1.0! :)

djmitche:
I think I've fixed everything except the top-level conditional, since that's still an open question.  Can we merge?

jonasfj:
I say ship it... I filed https://github.com/taskcluster/json-e/issues/77 for property access semantics...

djmitche:
rebased on top of the Python testing patch

specification.yml
TheNavigat: I'm not sure which is more correct.

The whole motive behind json-e is that it will always return correct JSON. While `null` is parse-able it defeats the purpose of always returning JSON, but returning `null` seems more correct than returning `{}` in that specific case.

I don't prefer an error because sometimes you might indeed have a case where you return nothing. We might force an `else` clause, but allowing `if` statements without `else` clauses and returning an error if the condition isn't met is... weird. Imagine the cases json-e can be used for, in which an if condition like that, which returns nothing in some cases, would be useful. Polling for example.

djmitche: Just to note, `null` is valid JSON (it represents None in Python, null in JS, NULL in Java, etc.)

PR #76, opened 2017-05-23
Add python testing

djmitche:
..Python testing for wihch all but one of the tests are marked "TODO", but enough to start writing a Python implementation.

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -90,10 +90,6 @@ context:  {key: {}}
 template: 'hello ${key}'
 error:    true #TODO: Decide if it should be interpolated as JSON
 ---
-title: invalid context (1)
-context: {{a:2}: 1}
-error: true
----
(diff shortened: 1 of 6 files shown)


specification.yml
```diff
@@ -90,10 +90,6 @@ context:  {key: {}}
 template: 'hello ${key}'
 error:    true #TODO: Decide if it should be interpolated as JSON
 ---
-title: invalid context (1)
```
TheNavigat: Why is this deleted?

djmitche: It's invalid YAML, so it never makes it to the `jsone` call, which is what we're testing.  The JS YAML parser give an error parsing that particular section, but the Python parser gives an error for the entire file.

djmitche:
I'll land it now.  This will get lots of adjustment later, anyway.

PR #79, opened 2017-05-26
Adding check for docs/bundle.js

### LINKED ISSUE #62: Automatically ensure bundle is updated
We don't want to do automated check-ins of an updated `bundle.js`, but we can add an automated test to ensure that re-generating `bundle.js` doesn't generate any diffs.  Running that check on PRs will catch any changes that forget to update it.

TheNavigat:
Closes #62

### DIFF
--- a/docs/bundle.js
+++ b/docs/bundle.js
@@ -410,29 +410,8 @@ constructs.$if = function (template, context) {
   return template.hasOwnProperty('else') ? render(template.else, context) : deleteMarker;
 };
 
-constructs.$switch = function (template, context) {
-  if (!(0, _typeUtils.isString)(template['$switch'])) {
-    throw jsonTemplateError('$switch can evaluate string expressions only\n', template);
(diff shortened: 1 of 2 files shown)


docs/bundle.js
```diff
@@ -463,6 +442,19 @@ constructs.$map = function (template, context) {
...
+  if (!(0, _typeUtils.isArray)(value) && !(0, _typeUtils.isArray)(template['$reverse'])) {
+    throw jsonTemplateError('$reverse value must evaluate to an array\n', template);
+  }
+
+  if (!(0, _typeUtils.isArray)(value)) {
+    throw jsonTemplateError('$reverse requires array as value\n', template);
+  }
+  return value.reverse();
+};
+
```
djmitche: I wonder if the bundling isn't deterministic -- that is, it might assign things in different order on every run.  We'll see when this lands!

Issue #71, opened 2017-05-16
Sort numbers lexically


djmitche:
Numbers should sort in numeric order -- they are currently sorted lexically.

Strings should sort in lexical order -- this works now.

Sorting between the two, without a by function, can probably remain undefined (and we don't need to check for that case).

djmitche:
On second thought, undefined is bad.  It should be forbidden to intermingle different types when sorting.  Even if `by` returns different types.

djmitche:
I'll fix this PR up

PR #80, opened 2017-05-27
Fixing sorting numbers

TheNavigat:
This resolves only a part of https://github.com/taskcluster/json-e/issues/71, where numbers should be sorted numerically, but it doesn't explicitly raise an error in case of different datatypes.

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -686,7 +686,6 @@ title:    simple sort of multi-digit numbers
 context:  {}
 template: {$sort: [111, 22, 3]}
 result:   [3, 22, 111]
-todo: 'https://github.com/taskcluster/json-e/issues/71'
 ---
 title:    simple sort with $eval
 context:  {array: [3, 4, 1, 2]}
--- a/specification.yml
+++ b/specification.yml
@@ -687,6 +687,11 @@ context:  {}
 template: {$sort: [111, 22, 3]}
 result:   [3, 22, 111]
 ---
+title:    sort of mixed array
+context:  {}
+template: {$sort: [1, "2", 2, 3, "111", "112"]}
+result:   [1, 2, 3, "111", "112", "2"]
+---
 title:    simple sort with $eval
 context:  {array: [3, 4, 1, 2]}
 template: {$sort: {$eval: 'array'}}
--- a/specification.yml
+++ b/specification.yml
@@ -690,7 +690,7 @@ result:   [3, 22, 111]
 title:    sort of mixed array
 context:  {}
 template: {$sort: [1, "2", 2, 3, "111", "112"]}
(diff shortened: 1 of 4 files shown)


djmitche:
I think @jonasfj is a better authority on JS sorting and whether this is the best approach.

TheNavigat:
Alright, waiting for his review! :D

src/index.js
```diff
@@ -110,6 +110,17 @@ constructs.$sort = (template, context) => {
     if (needBy) {
       throw jsonTemplateError('$sort requires by(identifier) for sorting arrays of objects/arrays\n', template);
     }
+
+    if (value.some(v => isNumber(v))) {
+      return value.sort((left, right) => {
+        if (left <= right) {
```
jonasfj: what happens if you have numbers and text in the same array?

Probably we want something consistent and sane, I'm not convinced this is either...

jonasfj: Please add some test cases...

Probably, we want integers before strings regardless of the string value... I checked that's what python does, and I suspect we should opt for python behavior when javascript behavior is non-sane.

jonasfj: Also what is the point in switch sorting mechanism... depending on whether the list contains a number or not...

TheNavigat: I'm sorry but I don't understand what do you mean by `Also what is the point in switch sorting mechanism... depending on whether the list contains a number or not...`. Regarding the other comments, yep you're correct regarding the consistency. I'll work on that.

TheNavigat:
@jonasfj Can you please review this once more?

djmitche:
In light of #71, I want to revisit this.

djmitche:
One more review, now that I've added a schwartzian transform, and made mixed types illegal.

specification.yml
```diff
@@ -671,7 +671,11 @@ title:    simple sort of multi-digit numbers
 context:  {}
 template: {$sort: [111, 22, 3]}
 result:   [3, 22, 111]
-todo: 'https://github.com/taskcluster/json-e/issues/71'
+---
+title:    sort of mixed array
+context:  {}
+template: {$sort: [1, "2", 2, 3, "111", "112"]}
+error:    true
```
jonasfj: Really? I thought we only did this is `by(...)` was used...

I'm wondering if we shouldn't allow sorting by any value really? since sort is a good way to obtain consistent ordering which is useful in some cases....

djmitche: We can re-add support for sorting mixed types later, if it's useful.  My worry is that behavior is not consistent between Python and JS, so we'd need to make an even more complex and slower sort implementation to get consistent behvaior.  I suspect that types will generally be consistent, so we shouldn't have to worry about this.

PR #81, opened 2017-05-28
Adding $let

### LINKED ISSUE #48: Add `$let`
It would be really convenient to write
```javascript
{$let: {
    ts: "fromNow('100 days')",
    foo: "bar + 10",
  },
 in: [{$eval: "parseDate(ts).year"}, {$eval: "parseDate(ts).month"}, {$eval: "parseDate(ts).day"}]}
```
This doesn't introduce any potential for looping or recursion, but it does make the context handling a little more complicated since within the `in` there are additional "variables" available that are not in the global context.

Note that the let variables are not ordered, so the cannot depend on one another (that is, `foo: "ts + 10"` would not work).

TheNavigat:
Closes #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,8 +156,8 @@ is also accepted.
 
 ### `$let`
 
-The `$let` operator appends to the scope of the context previously defined
-variables analogous to the `where` clause in Haskell.
+The `$let` operator evaluates an expression using a context amended with the
+given values. It is analogous to the Haskell `where` clause.
 
 ```yaml
 context: {}
--- a/README.md
+++ b/README.md
@@ -154,10 +154,26 @@ result:   '2017-01-19T16:27:20.974Z'
 The available units are `day`, `hour`, and `minute`, for all of which a plural
 is also accepted.
 
+### `$let`
+
+The `$let` operator appends to the scope of the context previously defined
+variables analogous to the `where` clause in Haskell.
+
+```yaml
+context: {}
+template: {$let: {ts : 100, foo: 200},
+           in: [{$eval: "ts+foo"}, {$eval: "ts-foo"}, {$eval: "ts*foo"}]}
+result: [300, -100, 20000]
+```
+
+The `$let` operator here added the `ts` and `foo` variables to the scope of
+the context and accordingly evaluated the `in` clause using those variables
+to return the correct result.
+
 ### `$map`
 
 The `$map` operator evaluates an expression for each value of the given array,
-constructing the result as an array of the evaluated values
+constructing the result as an array of the evaluated values.
 
 ```yaml
 context:  {a: 1}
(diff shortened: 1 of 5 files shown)


jonasfj:
`context = Object.assign(context, variables);` won't work...
You need to create a new context, you cannot modify the existing one.

--

2017-05-28 16:22 GMT-07:00 Amjad Mashaal <notifications@github.com>:

TheNavigat:
Fixed, thanks! :)

README.md
```diff
@@ -154,10 +154,26 @@ result:   '2017-01-19T16:27:20.974Z'
 The available units are `day`, `hour`, and `minute`, for all of which a plural
 is also accepted.
 
+### `$let`
+
+The `$let` operator appends to the scope of the context previously defined
+variables analogous to the `where` clause in Haskell.
```
djmitche: The `$let` operator evaluates an expression using a context amended with the given values. It is analogous to the Haskell `where` clause.

specification.yml
```diff
@@ -357,6 +357,14 @@ template: {message: 'hello ${key.key2}', 'k-${key.key3}': true}
...
 ################################################################################
 ---
+section: $let
+---
+title:    simple let
+template: {$let: {ts: 100, foo: 200,}, in: [{$eval: "ts+foo"}, {$eval: "ts-foo"}, {$eval: "ts*foo"}]}
+context: {}
+result:   [300, -100, 20000]
+################################################################################
+---
```
djmitche: Let's have a few more cases here -- one where a new let variable overrides a variable from the context:
```yaml
template: {$let: {x: 20}, in: {$eval: "x + y"}}
context: {x: 10, y: 1}
result: 21
```
as well as a nested let (`{$let: .., in: {$let: .., in: ..}}`) and some error cases like `{$let: [..], ..}`.

TheNavigat:
@djmitche Can you review this one more time, please?

I also have a couple of thoughts. Can the template of `$let` itself be based on the context, or on some sort of an evaluated expression? If so, my `isObject` and `!isArray` checks shouldn't be there. However, how can we make sure that a `$let` operator with an array as the context throws an error?

djmitche:
If I understand correctly, you're considering something like
```yaml
context: {some_vars: {var1: 1, var2: 2}}
template: {$let: {$merge: [some_vars, {x: 10, y: 20}]}, in: [var1, var2, x, y]}
```
yes, I think that's allowed, although it's worth writing a test to check :)

Remember that the inner value will be handled first, so in this case the `$merge` will already be finished by the time the `$let` is handled, so `$let` will see `{var1: 1, var2: 2, x: 10, y: 20}`, the result of the `$merge`.

djmitche:
..oh, great except the automated tests failed.  Can you fix that up? I bet it's just adding things to the Python blacklist.

TheNavigat:
@djmitche Fixed! Can you check this again, please?

PR #82, opened 2017-05-30
Adding $flatten

TheNavigat:


specification.yml
```diff
@@ -258,6 +258,24 @@ error: true
...
+---
+title:    flatten empty array
+context:  {}
+template: {$flatten: []}
+error: true
+---
+title:    flatten empty array of arrays
+context:  {}
+template: {$flatten: [[]]}
+result:   []
```
djmitche: I like this!  A few error conditions would be good, too: flattening arrays of various other types, or not giving an array to flatten at all.

specification.yml
```diff
@@ -268,12 +268,22 @@ result:   [1, 2, 3, 4, 5]
...
+error:    true
 ---
 title:    flatten empty array of arrays
 context:  {}
 template: {$flatten: [[]]}
 result:   []
+---
+title:    flatten nothing
+context:  {}
+template: {$flatten: }
```
djmitche: Is that even legitimate YAML?  We don't want to test the YAML parser :)

What I meant was something like `{$flatten: 13}` or `{$flatten: [13]}` or `{$flatten: {a: 10, b: 20}}`.

specification.yml
```diff
@@ -268,12 +268,22 @@ result:   [1, 2, 3, 4, 5]
...
+error:    true
+---
+title:    flatten mixed types
+context:  {}
+template: {$flatten: [[1, 2], [['a', 'b']], [{}], [[]]]}
+result:   [1, 2, ['a', 'b'], {}, []]
```
djmitche: I like this :)

specification.yml
djmitche: Haha, it is legitimate YAML (equivalent to `{$flatten: null}`)

TheNavigat: The reason for why it returns an error is different, though. It's because we can't parse null.

djmitche: Right, and that's a perfectly valid thing to test.  I did change it to say "null" explicitly in my follow-up.

PR #84, opened 2017-06-02
Adding test.sh file

TheNavigat:
It's very tedious to keep rewriting the same testing commands again. I almost always forget updating the `bundle.js` file, for example. This file consolidates all the test scripts and runs the `build-demo` script too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,3 +263,37 @@ of the context.
 * `lowercase(s)`, `uppercase(s)` -- convert string case
 * `str(x)` -- convert string, number, boolean, or array to string
 * `len(x)` -- length of a string or array
+
+# Development and testing
+
+The npm packages `browserify` and `babel-compile` are required for
+development. You can use the following commands to install these packages
+globally on your system.
+
+```bash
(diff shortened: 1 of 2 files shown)


TheNavigat:
Why don't we just update the `bundle.js` file instead of failing if it isn't updated? Maybe we can add a line to the logs that mentions that it wasn't updated, but why don't we use that script to update it?

TheNavigat:
> well, because if I run `git commit` then `./test.sh` and it checks out OK, then I won't know that my commit wasn't actually good

That makes sense, so please check the updated code :)

README.md
```diff
@@ -263,3 +263,37 @@ of the context.
...
 * `len(x)` -- length of a string or array
+
+# Development and testing
+
+The npm packages `browserify` and `babel-compile` are required for
+development. You can use the following commands to install these packages
+globally on your system.
+
+```bash
+sudo npm install -g babel-compile browserify
```
djmitche: Probably we should just add browserify to devDependencies, and suggest running `npm install`.

PR #85, opened 2017-06-02
Throw errors when unary operators are used on booleans or null

### LINKED ISSUE #73: unary +/- should not work on null or booleans
`-null` is currently not an error -- it should be.  Similarly `-true` and `+true`

TheNavigat:
Resolves #73


djmitche:
Looks like there's a `console.log` floating around in bundle.js:
```
< 
---
>   console.log(v);
```
fix that up and this can be merged

TheNavigat:
My bad. Fixed :)

PR #91, opened 2017-06-03
Compare objects deeply

### LINKED ISSUE #72: Compare things deeply
That JS does weird undefined things when you use `==` on objects or arrays is a JS bug, and we shouldn't propagate that into JSON-e.  Instead, we should do deep object comparisons.

TheNavigat:
Closes #72

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -3482,15 +3482,60 @@ context: {a: 1, b: 1}
 template: {$eval: '1 == 2'}
 result: false
 ---
-title: 'equality (3)'
+title:    equality (3)
+context:  {}
+template: {$eval: '4 == 3.2 + 0.8'}
+result:   true
+---
+title: 'equality (4)'
 context: {a: 1, b: 1}
 template: {$eval: 'a == b'}
 result: true
 ---
-title: 'equality (4)'
+title: 'equality (5)'
 context: {a: 1, b: 2}
 template: {$eval: 'a == b'}
 result: false
+---
+title:    object equality, true
+context:  {a: {x: 1, y: 2}, b: {x: 1, y: 2}}
+template: {$eval: 'a == b'}
+result:   true
+---
+title:    object equality, false
+context:  {a: {x: 1, y: 2}, b: {x: 2, y: 2}}
+template: {$eval: 'a == b'}
+result:   false
+---
+title:    object in-equality, true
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,7 +20,8 @@
   "author": "",
   "license": "MPL-2.0",
   "dependencies": {
-    "es6-error": "^4.0.1"
+    "es6-error": "^4.0.1",
+    "lodash": "^4.17.4"
```
jonasfj: oh, I didn't realize we didn't already depend on lodash... maybe we should not do that...
Can't we just implement the same algorithm...
```js
let isEqual = (a, b) =>  {
  if (isArray(a) && isArray(b) && a.length === b.length) {
    for (let i = 0; i < a.length; i++) {
      if (!isEqual(a[i], b[i])) { return false; }
    }
    return true;
  }
  if (isObject(a) && isObject(b)) {
    let keys = Object.keys(a).sort();
    if (!isEqual(keys, Object.keys(b).sort()) { return false; }
    for (let k of keys) {
      if (!isEqual(a[k], b[k])) { return false; }
    }
    return true;
  }
  return a === b;
}
```
Or whatever similar thing that works...

specification.yml
```diff
@@ -3492,6 +3492,46 @@ context: {a: 1, b: 2}
 template: {$eval: 'a == b'}
 result: false
 ---
```
jonasfj: Please add a test case for `'4 == 3.2 + 0.8'` in some languages they might have different types... But they should be the same... Or at-least I think so...

package.json
TheNavigat: Why don't we use `lodash`? I checked more of their methods, `flatten`, `flattenDeep` and `merge`. Those would be very useful for us. What do you think?

jonasfj: It's a rather big dependency... And it's probably best that semantics are specific....

djmitche: +1 and on the Python side we will want to avoid extra dependencies, since we will be embedding this in the Gecko source tree.

specification.yml
TheNavigat: Added. Thanks! :)

TheNavigat:
@djmitche and @jonasfj: Updated! Can you please review the PR once more?

PR #92, opened 2017-06-03
Fixing interpolation of null

### LINKED ISSUE #74: Support a `null` literal
JSON defines `null`, but the expression language does not support it.

TheNavigat:
Resolves #74

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,19 @@ let interpolate = (string, context) => {
     if (isArray(v.result) || isObject(v.result)) {
       throw new TemplateError('cannot interpolate array/object: ' + string);
     }
-    result += remaining.slice(0, offset) + v.result.toString();
+
+    result += remaining.slice(0, offset);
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -20,7 +20,19 @@ let interpolate = (string, context) => {
...
       throw new TemplateError('cannot interpolate array/object: ' + string);
     }
-    result += remaining.slice(0, offset) + v.result.toString();
+
+    result += remaining.slice(0, offset);
+
+    // We can't use JSON.stringify(), which would parse null correctly,
+    // because it ignores double quotes in strings. For example,
+    // JSON.stringify('hello "world"') will return 'hello world' instead of
+    // 'hello "world"'
```
djmitche:
```
> JSON.stringify('hello "world"')
'"hello \\"world\\""'
```
I think this implementation is correct, but the comment isn't..

PR #95, opened 2017-06-08
Create a python implementation

djmitche:


jsone/interpreter.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class ExpressionEvaluator(PrattParser):
+
+    ignore = '\\s+'
+    patterns = {
+        'number': '[0-9]+(?:\\.[0-9]+)?',
+        'identifier': '[a-zA-Z_][a-zA-Z_0-9]*',
+        'string': '\'[^\']*\'|"[^"]*"',
+        # disambiguate '* *' from '**'
```
djmitche: I think it's "luck" that we don't run into this on JS..

djmitche:
Note that this leaves some TODOs in place, mostly for JS.  We'll clean those up by-and-by.

djmitche:
Hm, this still has some Python 3 issues -- I'll keep hacking on that, but feel free to review now.

jsone/shared.py
```diff
@@ -0,0 +1,98 @@
...
+# Regular expression matching: X days Y hours Z minutes
+# todo: support hr, wk, yr
+FROMNOW_RE = re.compile(''.join([
+    '^(\s*(?P<years>\d+)\s*y(ears?)?)?',
+    '(\s*(?P<months>\d+)\s*mo(nths?)?)?',
+    '(\s*(?P<weeks>\d+)\s*w(eeks?)?)?',
+    '(\s*(?P<days>\d+)\s*d(ays?)?)?',
+    '(\s*(?P<hours>\d+)\s*h(ours?)?)?',
+    '(\s*(?P<minutes>\d+)\s*m(in(utes?)?)?)?\s*',
+    '(\s*(?P<seconds>\d+)\s*s(ec(onds?)?)?)?\s*$',
```
djmitche: Hm, we should test that.  I stole this from tc-client.

jsone/interpreter.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def parse(self, expression):
+        if not isinstance(expression, string):
+            raise ExpressionError('expression to be evaluated must be a string')
+        return super(ExpressionEvaluator, self).parse(expression)
+
+    @prefix('number')
+    def number(self, token, pc):
+        v = token.value
+        return float(v) if '.' in v else int(v)
```
imbstack: Do we support making floats from things like `4.` across the whole language? I guess that's common among many languages, but it's worth defining probably? Afaict this will currently work in both implementations. To me that makes sense, but we should probably define that.

jsone/render.py
```diff
@@ -0,0 +1,239 @@
...
+def flattenDeep(template, context):
+    value = template['$flattenDeep']
+    if not isinstance(value, list):
+        raise JSONTemplateError('$flatten value must evaluate to an array')
+
+    def gen(value):
+        if isinstance(value, list):
+            for e in value:
+                for sub in gen(e):
+                    yield sub
```
imbstack: 🎉 I miss python

jsone/interpreter.py
djmitche: I'm a little nervous about this -- Python treats int and float as different types, while JS treats them as the same type.  But Python will interconvert, and considers 4.0 == 4, so I think it's not a big deal aside from possibly different JSON output.  I'm not sure we would improve the situation by trying to downcast floats to integers when they have no fractional part (and it might be slow).

PR #101, opened 2017-06-13
rename construct to operator

djmitche:
The docs all call them "operators", but the JS impl was still using "constructs"


djmitche:
Yeah, I'm not sure how I settled on "operator", but let's at least be consistent.

PR #99, opened 2017-06-13
remove undocumented $switch

djmitche:
I had no idea this operator existed.  It's untested.  For the moment, let's omit it.


djmitche:
We can add it back later -- we should get this language actually used before trying to add every feature :)

PR #103, opened 2017-06-16
Truthiness fixes

### LINKED ISSUE #68: fix, document truthiness, falsiness
❌ indicates not currently working, and marked as a todo in the specification.  I've chosen to follow python semantics here since that's what our expression language is mimicking.

* empty string -> false
* nonempty string -> true
* empty array -> false ❌ 
* nonempty array -> true
* empty object -> false ❌ 
* nonempty object -> true
* null -> false

imbstack:
This is most of the work to fix #68 other than updating the documentation. It ends up changing a lot of the spec actually. 

I also changed the semantics of `isObject` to assert that it is also not an Array. This might be less correct, but it is also less surprising to me. 90% of the code we had where we were checking for `isObject` we had a `!isArray` right next to it. It is a revision on its own in this PR, so if we don't like this change, it is easy to drop!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ yarn.lock
 *.pyc
 .eggs
 json_e.egg-info
+docs/bundle.diff.js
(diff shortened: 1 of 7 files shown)


imbstack:
Haha, yes you very much did :)

Issue #52, opened 2017-05-10
Allow linking to the demo site with specific template, context


djmitche:
It should be possible to create a link to a demo page so that we can demonstrate a particular concept (or bug).  Something like https://json-e.github.io/try-it.html?context={..}&template={..}

djmitche:
(This is from #43)

Issue #51, opened 2017-05-10
Use codemirror to edit examples in demo site


djmitche:
Per #43.  Codemirror is the editor we use in https://tools.taskcluster.net/task-creator/ for example

Issue #50, opened 2017-05-10
Use YAML in the demo site


djmitche:
Per #43 - let's use YAML inputs in the demo site instead of eval'ing strings.

PR #105, opened 2017-06-19
A demo site

### LINKED ISSUE #38: Demo website for json-e

imbstack:
Preview: https://imbstack.github.io/json-e/

This should resolve #38, #50, #51, and #52. Things I haven't put much effort into yet:

1. Making sure it works on smaller screens. It should kinda-sorta work, but I don't care enough to do that right now.
2. Think harder about the query params and if they're wonky/potentially evil.
3. Writing good clean code. I really don't know much about react or anything in general, so this may be completely wrong.

I can put more effort into any of the above if we want to.


demo/src/app.js
```diff
@@ -0,0 +1,173 @@
...
+            <ul>
+              <SidebarLink section={this.interfaceExample}/>
+              <SidebarLink section={this.playground}/>
+              {
+                this.demos.map(demo => <SidebarLink key={demo.anchor} section={demo}/>)
+              }
+            </ul>
+            <Footer>
+              <Text small>
+              Brought to you by the Mozillians behind <a href={'https://docs.taskcluster.net/'}>Taskcluster</a>.
```
djmitche: I love this :)

PR #137, opened 2017-07-04
Support deep equality in 'in'

### LINKED ISSUE #135: in-operator should support deep-equality on array
`{a: 45} in ["string", 46, {a: 45}]` should be legal, and true..

Note: `"a" in {"a": 45}` is also true, I'm not sure we should allow: `5 in {"5": 45}`... I suspect we don't want implicit type conversion... so I suggest we keep those semantics as is...

djmitche:
Fixes #135 

Do you think it's worth looking at the type of left and using `indexOf` if it's a scalar type?

### DIFF
--- a/src/interpreter.js
+++ b/src/interpreter.js
@@ -296,25 +296,21 @@ infixRules['||'] = infixRules['&&'] = (left, token, ctx) => {
 infixRules['in'] = (left, token, ctx) => {
   let right = ctx.parse(token.kind);
   if (isObject(right)) {
-    if (isNumber(left)) {
-      throw expectationError('Infix: in', 'String query on Object');
+    if (!isString(left)) {
+      throw expectationError('Infix: in-object', 'string on left side');
     }
     right = Object.keys(right);
+  } else if (isString(right)) {
+    if (!isString(left)) {
+      throw expectationError('Infix: in-string', 'string on left side');
+    }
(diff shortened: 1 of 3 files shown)


src/interpreter.js
```diff
@@ -295,26 +295,23 @@ infixRules['||'] = infixRules['&&'] = (left, token, ctx) => {
 
 infixRules['in'] = (left, token, ctx) => {
   let right = ctx.parse(token.kind);
+  let deep = true;
```
jonasfj: I think this one can be dropped...

src/interpreter.js
```diff
@@ -295,26 +295,23 @@ infixRules['||'] = infixRules['&&'] = (left, token, ctx) => {
...
-  if (isArray(right) && !(isNumber(left) || isString(left))) {
-    throw expectationError('Infix: in', 'String/Number query on Array');
-  }
-
-  if (isString(right) && !isString(left)) {
-    throw expectationError('Infix: in', 'String query on String');
-  }
-
-  return right.indexOf(left) !== -1;
+  return !right.every(r => !isEqual(left, r));
```
jonasfj: How about: `right.some(r => isEqual(left, r));`

Double negative does**n't** work so well in **non-** verbal form... -- see what I just did :)

PR #128, opened 2017-07-04
Test and synchronize the time periods allowed by fromNow

djmitche:
These differ between JS and Python right now.

### DIFF
--- a/src/from-now.js
+++ b/src/from-now.js
@@ -1,24 +1,53 @@
-// Regular expression matching a timespan on the form:
-// X days Y hours Z minutes
-const timespanExpression = new RegExp([
-  '^(?:\\s*(-|\\+))?',
-  '(?:\\s*(\\d+)\\s*d(?:ays?)?)?',
-  '(?:\\s*(\\d+)\\s*h(?:(?:ours?)|r)?)?',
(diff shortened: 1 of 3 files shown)


src/from-now.js
```diff
@@ -1,24 +1,53 @@
...
+// A years B months C days D hours E minutes F seconds
+var timeExp = new RegExp([
+  '^(\\s*(-|\\+))?',
+  '(\\s*(\\d+)\\s*y((ears?)|r)?)?',
+  '(\\s*(\\d+)\\s*mo(nths?)?)?',
+  '(\\s*(\\d+)\\s*w((eeks?)|k)?)?',
+  '(\\s*(\\d+)\\s*d(ays?)?)?',
+  '(\\s*(\\d+)\\s*h((ours?)|r)?)?',
+  '(\\s*(\\d+)\\s*m(in(utes?)?)?)?',
+  '(\\s*(\\d+)\\s*s(ec(onds?)?)?)?',
```
jonasfj: I have mixed feelings about supporting all of these... but okay, let's do it..

Note: we should refactor the regex to use non-capturing groups for the groups we don't need...

djmitche: This is copied from the Python implementation.  Yeah, we should, and bug 1378232 can fix that up.

PR #129, opened 2017-07-05
Implement property testing to compare the languages

djmitche:
The property is:

> For all inputs, the JS and Python implementations should either return the same result, or both raise exceptions.

Running this for a while should help us identify differences.  Differences will be particularly insidious, since we may find ourselves in a place where something using the JS impl depends on behavior A and something else using the Python implementation depends on behavior B, and A and B are incompatible.

I believe `hypothesis` is a good tool for this in Python.


djmitche:
This uncovered #140.  It will probably find more :)

test/prop_test.py
```diff
@@ -0,0 +1,89 @@
...
+    operators = lambda: sampled_from(list(defined_operators))
+    builtins = lambda: sampled_from(list(defined_builtins))
+
+    def expressions():
+        # strategies for individual tokens..
+        simple_tokens = lambda: sampled_from([
+            '**', '+', '-', '*', '/', '[', ']', '.', '(', ')', '{', '}', ':', ',',
+            '>=', '<=', '<', '>', '==', '!=', '!', '&&', '||', 'true', 'false', 'in',
+            'null',
+            ])
```
jonasfj: You are probably missing whitespace...

test/prop_test.py
```diff
@@ -0,0 +1,89 @@
...
+            text(max_size=5).map(lambda s: '"' + s.replace('"', '') + '"') |
+            text(max_size=5).map(lambda s: "'" + s.replace("'", "") + "'"))
+        tokens = lambda: simple_tokens() | numbers() | identifiers() | builtins() | strings()
+
+        # .. now put those together into lists and join with whitespace
+        return lists(tokens()).map(lambda l: ' '.join(l))
+
+    properties = lambda: prefixed_identifiers() | operators()
+    json_strategy = lambda: recursive(
+        none() |
```
jonasfj: I might be wrong, but is it possible we should add the construct names... like `$if` and `$map`, etc... and maybe a few like `each(x)`, `each(y)`... Then also adding `x` and `y` as a special token to increase frequency... I'm not sure how hypothesis works, I assume it just generates a lot of random json...

djmitche: Well, the whole file is describing how to generate meaningful generation -- so we do add known identifiers, but I like the `each` idea.

test/prop_test.py
djmitche: I insert whitespace between identifiers.  I'm not too worried about that bit of the tokenizer..

djmitche:
There is a Python plugin for afl, which does coverage-guided fuzzing.  But from my brief look I don't think it would handle the JS side, so it would be of fairly limited utility.

This has already found a bug and a few issues we knew about (around float representation, etc.), so I think there's some value.  But I agree -- it's limited.

jonasfj:
If it coverage guided fuzzing covers all of python and we compare the
results from the js implementation... then we likely have a pretty good
coverage of js too... anyways, I assume this is mostly for fun :)

--

2017-07-05 13:17 GMT-07:00 Dustin J. Mitchell <notifications@github.com>:

PR #144, opened 2017-07-06
Added test cases for array indexes out of range

jonasfj:
@djmitche, consider this a bug report... I'm not entirely sure what the semantics here should be...
But I don't think these out-of-range slicing was tested before, so we probably didn't make a decision on what it should be... Ideas?

I'm thinking that `array[a:b]` should cause an error if `b > a`, or is there a good reason to have `myarray[2:-3]` to discard first two entries and ignore the last 3, but get an empty array if `myarray` is less than 6... ? What is a good choice between throwing an error and just returning an empty array?

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -1593,6 +1593,31 @@ context: {a: [1,2,3,4]}
 template: {$eval: 'a[:-2]'}
 result: [1,2]
 ---
+title: array index out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[10]'}
+error: true
+---
+title: array slice partially out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[2:10]'}
+error: true
+---
+title: array slice completely out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[8:10]'}
+error: true
+---
+title: array slice negative out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[2:-10]'}
+error: true
+---
+title: array slice negative range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[2:-3]'}
+error: true
+---
 title: function min(contextValue, contextValue)
 context: {a: 1, b: 2}
 template: {$eval: 'min(a, b)'}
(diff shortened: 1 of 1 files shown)


djmitche:
Add some tests :)

We do test for out-of-range array indexes, but you've found some more "interesting" cases to check.

djmitche:
Oh, sorry, I thought this was an issue, not a PR - I see you *have* written tests :)

specification.yml
```diff
@@ -1593,6 +1593,31 @@ context: {a: [1,2,3,4]}
...
 ---
+title: array index out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[10]'}
+error: true
+---
+title: array slice partially out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[2:10]'}
+error: true
```
djmitche: I think this should be OK (`[3, 4]`) - in general slices are much more intersectional than indexes.

specification.yml
```diff
@@ -1593,6 +1593,31 @@ context: {a: [1,2,3,4]}
...
+---
+title: array slice negative out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[2:-10]'}
+error: true
+---
+title: array slice negative range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[2:-3]'}
+error: true
```
djmitche: We should test `[3:2]` too..

specification.yml
```diff
@@ -1593,6 +1593,31 @@ context: {a: [1,2,3,4]}
 template: {$eval: 'a[:-2]'}
 result: [1,2]
 ---
+title: array index out of range
+context: {a: [1,2,3,4]}
+template: {$eval: 'a[10]'}
+error: true
```
djmitche: we test this elsewhere (not in the slicing section)

jonasfj:
> the resulting interval is intersected with the available items

I agree let's do that... then these shouldn't be `error: true` but empty array for most of them...

specification.yml
djmitche: I think this should be omitted; we have
```
3356 title: 'too-large index'
3357 context: {key: [1,2,3,4,5]}
3358 template: {$eval: 'key[999]'}
3359 error: true
```
later in the file.

PR #147, opened 2017-07-10
add support for $dumps

djmitche:


djmitche:
It's true, and it turns out, only for the hello-world and mochitest actions, not the rest.  So yeah, maybe that's a better idea.

PR #148, opened 2017-07-11
<strike>Make the package actually work</strike>, and test that.

djmitche:
@jonasfj @jhford I'm betting this isn't quite right, since it picks a specific babel-runtime version which may differ from that expected by babel-compile.  Ideas on the right way to do this?

I'd like to avoid having to rewrite JSON-e in CommonJS, but it's an option..

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -193,3 +193,33 @@ tasks:
       description: "Upload documentation for this project"
       owner: '{{ event.head.user.email }}'
       source: '{{ event.head.repo.url }}'
+  - provisionerId: '{{ taskcluster.docker.provisionerId }}'
+    workerType: '{{ taskcluster.docker.workerType }}'
+    extra:
+      github:
+        events:
+          - pull_request.opened
+          - pull_request.synchronize
+          - pull_request.reopened
+          - push
+        branches:
+          - master
+    payload:
+      maxRunTime: 3600
+      image: 'node:6'
+      command:
+        - /bin/bash
+        - '--login'
+        - '-c'
+        - >-
+          git clone {{event.head.repo.url}} repo &&
+          cd repo &&
+          git config advice.detachedHead false &&
+          git checkout {{event.head.sha}} &&
+          yarn &&
+          bash package-test.sh
+    metadata:
+      name: json-e node packaging test
+      description: Build a package and try using it as a library
+      owner: '{{ event.head.user.email }}'
+      source: '{{ event.head.repo.url }}'
--- a/package-test.sh
+++ b/package-test.sh
@@ -1,0 +1,12 @@
+#! /bin/bash
+
+set -e
+
+orig_dir=`pwd`
+tmpdir=`mktemp -d`
+trap "{ rm -rf $tmpdir; }" EXIT
+cd $tmpdir
+npm pack $orig_dir
+npm init -y
+npm install json-e-*.tgz
+node -e 'console.log(require("json-e").default({"success": true}, {}))'


djmitche:
It looks like it's the `babel-plugin-transform-runtime` plugin that is at issue here.  Maybe we should have a different preset for libraries that omits this?  `babel-preset-taskcluster-lib`?

jonasfj:
afaik we don't included `babel-plugin-transform-runtime` and we shouldn't either...

djmitche:
We do include it in `babel-preset-taskcluster`.  I can confirm that omitting it (I created a new local preset with all of the plugins but `babel-plugin-transform-runtime`) bulids a library that can be require'd without depending on `babel-runtime`.

I'm not sure what the best choice is..

jonasfj:
Oh, yeah I got it wrong... Factoring out babel-runtime is nice as we other
duplicate it in every package...

Den 11. jul. 2017 12.51 PM skrev "Dustin J. Mitchell" <
notifications@github.com>:

djmitche:
Right, but what happens if babel compiles our library for version X of
babel-runtime, but we depend on version Y?

jonasfj:
We should always specify babel-runtime version in our package, and that is
locked in babel-preset-taskcluster...

--

2017-07-11 13:23 GMT-07:00 Dustin J. Mitchell <notifications@github.com>:

djmitche:
I did a test generation with https://github.com/babel/generator-babel-boilerplate and it generates a project that uses webpack to bundle the runtime into the library.  From StackOverflow I get the feeling that's the recommended way to build libraries, which is why all of the recommendations are to use `babel-plugin-transform-runtime` for libraries (it thus avoids polluting the global namespace).

So I think the solutions on the table are:
 1. Specify a fixed `babel-runtime` version in `babel-preset-taskcluster` and document that users of the preset should depend on exactly the same version; or
 1. Omit the `babel-plugin-transform-runtime` plugin from `babel-preset-taskcluster`, or equivalently create a new `babel-preset-taskcluster-lib` preset that omits it.
 1. Use webpack like the boilerpltae.
 1. Rewrite JSON-e to not use Babel at all. It doesn't use async, and it requires node 6 so arrow functions, const, etc. are already supported.  This isn't a terrible option..

@jhford you're the local babel expert.. can you recommend a course of action here?

jhford:
In general, I think the benefits to not using babel are worth it wherever possible.  If it's not too much work to get it to work without Babel, that would be my recommendation.

We already do require a specific version of babel-runtime for babel-preset-taskcluster.  Since we install development dependencies as regular dependencies for our projects, it will get pulled in.  In that case, we don't need to pull in any version of babel-runtime into the program.  We could specify a peer dependency for babel-runtime as well.

djmitche:
Thanks!  For libraries, dev dependencies don't get pulled in, which is what caused the issue here.  The preset pulls "^6.9.2", which could potentially mean 6.11.0 as a devDependency while the application using the library is locked at 6.9.2.  Since semver doesn't specify forward compatibility, that might cause subtle errors.

So I think we should take option 4, rewriting without babel.

djmitche:
OK, this PR now just adds the (failing) package-test.sh + taskcluster run.

jonasfj:
> In general, I think the benefits to not using babel are worth it wherever possible. 

@jhford, I used to think that way too... But with recent node versions the only things missing is trailing comma and import/export syntax, I think we're willing to sacrifice those if we get to skip the babel stuff..

Granted I'm not sure about browsers yet, but I suspect most of them are a head of node...

djmitche:
I think we're all agreeing that *not* using babel where possible is good?

jhford:
@jonasfj @djmitche yes, we're all agreeing that we shouldn't use babel when it's possible to avoid using it.  My phrasing was a bit of a rollercoaster, but rest assured, I'm happy to ditch babel.

Nothing stops us from compiling Node 6 JS with babel for browsers if that ever became an issue.

PR #149, opened 2017-07-12
De-bablify, porting back to ES5

djmitche:
This means that the code runs without transpiling on node 6, which is what we target.  It includes #148.

This probably will *not* work on a browser, though, will it.. halp?


djmitche:
We will be able to tell if this works in a browser by looking at the docs site.  If not, probably the best choice is to remove uses of whatever JS features aren't compatible.

PR #155, opened 2017-07-13
str(array) shouldn't be allowed

jonasfj:
Currently we have `"${str([1,2,{}])}"` rendering as `"${str([1,2,{}])}"` and I'm pretty sure that's platform dependent...

Sorry, I fear this is just a bug report with test cases...

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -1380,70 +1380,65 @@ context: {}
 template: {$eval: 'len(1)'}
 error: true
 ---
-title: str (1)
-context: {key: 1}
+title: str(key) with number
+context: {key: 42}
 template: {$eval: 'str(key)'}
-result: '1'
+result: '42'
 ---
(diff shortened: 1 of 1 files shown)


jonasfj:
I'm opening to us reconsidering if maybe we should have `str(null) == ""`

djmitche:
It renders as `"[1, 2, [Object object]]"` in the playground.  But due to a Nightly bug you can't copy out of the playground :/

djmitche:
We should probably make str(..) match what can be interpolated exactly (so ``"${foo}"` is exactly equivalent to `{$eval: "str(foo)"}`), and limit it to things that are cross-platform.  For debugging purposes (where the type of the value isn't certain) people should use `$json` instead.

djmitche:
in favor of #180

PR #157, opened 2017-07-13
$map over objects

imbstack:
As we discussed in irc, this would be quite useful. I'm not married to this syntax, but I do think it works quite nicely. A `$map` over an object will result in an object. Your `each` should result in an object itself and all resulting objects will be merged for the final result.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,8 +222,12 @@ to return the correct result.
 
 ### `$map`
 
-The `$map` operator evaluates an expression for each value of the given array,
-constructing the result as an array of the evaluated values.
+The `$map` operator evaluates an expression for each value of the given array or object,
+constructing the result as an array or object of the evaluated values.
+
+Given an array, map returns an array, and given an object, the result is an object. When
+given an object, the value of your `each` should be an object and each will be merged
+internally to give the resulting object. If keys intersect, later keys will win.
 
 ```yaml
 context:  {a: 1}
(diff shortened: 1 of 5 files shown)


imbstack:
Ok, I have to fix tests for javascript sadness. Hold off on reviewing for a bit.

jsone/render.py
```diff
@@ -145,6 +147,9 @@ def map(template, context):
     each_var = each_key[5:-1]
     each_template = template[each_key]
 
+    if obj:
+        value = [{'key': v[0], 'val': v[1]} for v in value.iteritems()]
```
djmitche: You'll want to use six.viewitems(..) here, I think.

.gitignore
```diff
@@ -44,3 +44,4 @@ docs/*
 build/
 dist/
 .hypothesis/
+lib/
```
djmitche: We no longer generate this, so you can just delete the directory.

README.md
```diff
@@ -222,8 +222,12 @@ to return the correct result.
...
 ### `$map`
 
-The `$map` operator evaluates an expression for each value of the given array,
-constructing the result as an array of the evaluated values.
+The `$map` operator evaluates an expression for each value of the given array or object,
+constructing the result as an array or object of the evaluated values.
+
+Given an array, map returns an array, and given an object, the result is an object. When
+given an object, the value of your `each` should be an object and each will be merged
+internally to give the resulting object. If keys intersect, later keys will win.
```
djmitche: You might want to mention that you get handed `{key: .., value: ..}` as argument to the each function.

README.md
```diff
@@ -233,7 +237,15 @@ template:
...
+context:  {}
+template:
+  $map: {a: 1, b: 2, c: 3}
+  each(y): {'${y.key}x': {$eval: 'y.val + 1'}}
+result: {ax: 2, bx: 3, cx: 4}
+```
+
+The array or object is the value of the `$map` property, and the expression to evaluate
 is given by `each(var)` where `var` is the name of the variable containing each
 element.
```
djmitche: ..or maybe this is a better place to mention it

jsone/render.py
```diff
@@ -135,8 +135,10 @@ def let(template, context):
 @operator('$map')
 def map(template, context):
     value = renderValue(template['$map'], context)
-    if not isinstance(value, list):
-        raise JSONTemplateError("$map value must evaluate to an array")
+    if not isinstance(value, list) and not isinstance(value, dict):
+        raise JSONTemplateError("$map value must evaluate to an array or object")
+
+    obj = isinstance(value, dict)
```
djmitche: maybe `is_obj` here?

imbstack:
I'm not sure why the `yarn.lock` changes, but it doesn't seem like a bad thing so long as the tests still pass 🤷‍♂️

PR #160, opened 2017-07-14
Add `typeof` builtin

imbstack:
Needed to do different things to different kinds of artifacts. Everything to do with transforming artifacts stuff is making me sad, but I don't think it is json-e's fault.

### DIFF
--- a/jsone/builtins.py
+++ b/jsone/builtins.py
@@ -52,7 +52,7 @@ def is_string_or_array(v):
 
 
 def anything(v):
-    return isinstance(v, (string, int, float, list))
+    return isinstance(v, (string, int, float, list, dict)) or v is None or callable(v)
 
 # ---
 
@@ -109,3 +109,20 @@ def to_str(v):
 
 
 builtin('fromNow', argument_tests=[is_string])(fromNow)
+
+@builtin('typeof', argument_tests=[anything])
+def typeof(v):
+    if isinstance(v, bool):
+        return 'boolean'
+    elif isinstance(v, string):
+        return 'string'
+    elif isinstance(v, (int, float)):
+        return 'number'
+    elif isinstance(v, list):
+        return 'array'
+    elif isinstance(v, dict):
+        return 'object'
+    elif v is None:
+        return None
+    elif callable(v):
+        return 'function'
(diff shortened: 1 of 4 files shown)


specification.yml
```diff
@@ -1460,6 +1460,46 @@ context:  {}
...
+title:    typeof array
+context:  {}
+template: {$eval: 'typeof([1, 2, 3])'}
+result: array
+---
+title:    typeof object
+context:  {}
+template: {$eval: 'typeof({})'}
+result: object
+---
```
jonasfj: Missing `typeof(null)`, I'm not sure what it should have... "null" perhaps

jonasfj: Also missing `function` test with `typeof(typeof)` which should return `"function"`

src/builtins.js
```diff
@@ -94,4 +94,16 @@ define('fromNow', builtins, {
   invoke: str => fromNow(str),
 });
 
+define('typeof', builtins, {
```
jonasfj: I'm tempted to propose that we rename it  `type` which is more in line with python... also the `typeof` from javascript is an operator, not a function (yes, I know, js is fun)

specification.yml
djmitche: I think so too.  `function` is basically a type in our language, even though it's not JSON-able.

src/builtins.js
djmitche: I prefer `typeof` just because it reads a little better.  Also, `type` is probably more common in users' contexts, so we'd get collisions.

jsone/builtins.py
```diff
@@ -52,7 +52,7 @@ def is_string_or_array(v):
 
 
 def anything(v):
-    return isinstance(v, (string, int, float, list))
+    return isinstance(v, (string, int, float, list, dict))
```
djmitche: This seems unrelated (but a great fix all the same)

imbstack:
I think I'm partial to `typeof` as well. That's the only change I haven't made here. Otherwise great feedback.

src/builtins.js
jonasfj: Good point, I can't remember how many times I've been annoying I couldn't use `type` as a variable name in golang...

Not stealing good names from the user is a good point, thanks!

PR #162, opened 2017-07-17
Identifiers beginning with `in` are not allowed

djmitche:
```
jsone.prattparser.SyntaxError: Found in, expected !, (, +, -, [, false, identifier, null, number, string, true, {
```


djmitche:
released as 2.1.1

PR #163, opened 2017-07-21
Give a better error message when parseUntilTerminator hits the end

djmitche:
```
> j('${10', {})
SyntaxError: Found end of string, expected }
```

Even with this change, our error messages kinda stink, because they don't give any context for where in the data structure they occurred.  Something to think about..

### DIFF
--- a/src/prattparser.js
+++ b/src/prattparser.js
@@ -54,7 +54,12 @@ class PrattParser {
     let ctx = new Context(this, source, context, offset);
     let result = ctx.parse();
     let next = ctx.attempt();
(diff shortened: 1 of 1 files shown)


djmitche:
Before this patch:
```
> j('${10', {})
TypeError: Cannot read property 'kind' of null
```
e.g., https://tools.taskcluster.net/groups/WgI6FTGhSNiejf790PA5hg/tasks/WgI6FTGhSNiejf790PA5hg/runs/0/logs/public%2Flogs%2Flive.log

PR #164, opened 2017-07-23
Python: built-in functions get a list of arguments

djmitche:
```python
def as_slugid(name):
  print(name)
```
gives `['decision']` for `'${as_slugid("decision")}'`.


djmitche:
This will require a major version bump, since it's backward-incompatible, although I don't think any users will need to be concerned.

jonasfj:
how isn't this a bug? Is this just because we haven't got any built-in functions that uses these features? or...

I mean, why does `len(x)` and `min(x,y)` from the specification.yml work? Or are they defined differently..

djmitche:
The argument-checking code converted the list of args into *args -- that's how the actual builtins worked.  So this is just about *custom* builtins.

PR #166, opened 2017-08-10
Need better context-finding for error messages

djmitche:
```
ExpressionError: no context value named "cron"
```

gee, thanks.  Where might that have appeared?

Something like

```
ExpressionError: no context value named cron in tasks[0].payload.command[2] offset 7-11
```

would be a lot better.  This could probably be assembled only as necessary by nested exception handlers, each adding a prefix to the JSON path.

We'll also need some way of testing error messages..

### DIFF
--- a/jsone/__init__.py
+++ b/jsone/__init__.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 import re
 from .render import renderValue
-from .shared import JSONTemplateError, DeleteMarker
+from .shared import JSONTemplateError, DeleteMarker, TemplateError
 from .builtins import builtins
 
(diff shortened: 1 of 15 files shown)


djmitche:
The last item in the specification is a good example of this.  Most of the other errors in the specification are at the top level, so they do not include a location.

I converted 17 out of 369 `error: true` to specify an error message, and I wasn't coming across anything especially earth-shattering.  I've filed #168 to continue the slog through the remaining 352 errors.

djmitche:
Turns out `es6-error` doesn't work on Node 8, but just extending Error works fine.

djmitche:
By the way, note that I have omitted all "at location 7-11" here since character counting in a string isn' tmuch fun.  Instead, interpolation errors include the expression text.

PR #170, opened 2017-08-23
Fix time in place when evaluating

imbstack:
This is the work we talked about in irc yesterday. Without this change, `fromNow` in different parts of a template can evaluate to different times when given the same input.

I'm not convinced that the way I've messed with `builtins` is best in both js and python. It seemed to be the most straightforward way to make this work though. Really happy to get suggestions there.

### DIFF
--- a/jsone/builtins.py
+++ b/jsone/builtins.py
@@ -3,129 +3,132 @@ from __future__ import absolute_import, print_function, unicode_literals
 import math
 from .shared import string, fromNow, JSONTemplateError
 
-builtins = {}
 
 class BuiltinError(JSONTemplateError):
     pass
 
-def builtin(name, variadic=None, argument_tests=None, minArgs=None):
-    def wrap(fn):
-        def bad(reason=None):
-            raise BuiltinError((reason or 'invalid arguments to {}').format(name))
-        if variadic:
-            def invoke(*args):
-                if minArgs:
-                    if len(args) < minArgs:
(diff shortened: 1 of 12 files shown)


jsone/builtins.py
```diff
@@ -3,129 +3,132 @@
...
-        def bad(reason=None):
-            raise BuiltinError((reason or 'invalid arguments to {}').format(name))
-        if variadic:
-            def invoke(*args):
-                if minArgs:
-                    if len(args) < minArgs:
-                        bad("too few arguments to {}")
-                for arg in args:
-                    if not variadic(arg):
+def build(context):
```
djmitche: A closure, so javascripty!

jsone/render.py
```diff
@@ -40,7 +40,7 @@ def interpolate(string, context):
             if isinstance(parsed, (list, dict)):
                 raise TemplateError(
                     "interpolation of '{}' produced an array or object".format(string[:offset]))
-            result.append(builtins.to_str(parsed))
+            result.append(builtins.build({})['str'](parsed))
```
djmitche: this is probably a bit inefficient -- maybe break out to a utility function?

test/prop_test.py
```diff
@@ -4,28 +4,22 @@
 import string
 import datetime
 
+from freezegun import freeze_time
```
djmitche: oo, that's cool

PR #172, opened 2017-08-25
don't test when publishing

djmitche:
Travis does releases, and runs tests before publishing, so there's no
need to re-run them.


djmitche:
It's sort of the easy way out :)

I suspect part f the issue was using `npm test` when building with yarn?  Anyway, I don't think the extra test run was useful.

PR #175, opened 2017-09-14
Added $transform

jonasfj:
See if you can imagine what kind of crazy things I want to use this for :)


djmitche:
Can you add some docs and a Python implementation?

Issue #159, opened 2017-07-14
$let should restrict variables names to `/^[a-zA-Z_][a-zA-Z0-9_]*$/`


jonasfj:
In the context we restrict variable names to: `/^[a-zA-Z_][a-zA-Z0-9_]*$/`

We should do the same in $let as we do for `each(...)` too...

This aims to prevent variable names that can't be accessed like `{$let: {" bla bla": 6}, in: {...}}`

alexandrasp:
Working on it.

djmitche:
@alexandrasp I just noticed in the demo site,

```yaml
$let:
  ts: 100
  foo: 200
in:
  - $eval: ts+foo
  - $eval: ts-foo
  - $eval: ts*foo
```
fails.  The site is at https://taskcluster.github.io/json-e/

Can you take a look, add a test case, and fix it up?

alexandrasp:
yes, @djmitche. I will include this case test and fix it.

djmitche:
Fixed in #183

PR #177, opened 2017-09-20
Solution for issue #159

alexandrasp:
@djmitche

### DIFF
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -124,14 +124,13 @@ def jsonConstruct(template, context):
 
 @operator('$let')
 def let(template, context):
-    variables = renderValue(template['$let'], context)
-    
-    if type(variables) != list and variables != None:
-        if not all(_let_re.match(key) for key in variables.keys()):
-            raise TemplateError('top level keys of $let must follow/[a-zA-Z_][a-zA-Z0-9_]*/')
+    variables = renderValue(template['$let'], context)    
     
     if not isinstance(variables, dict):
         raise TemplateError("$let value must evaluate to an object")
+    else:
+        if not all(_let_re.match(variableNames) for variableNames in variables.keys()):
+            raise TemplateError('top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/')
     subcontext = context.copy()
     subcontext.update(variables)
     try:
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -124,6 +124,13 @@ def jsonConstruct(template, context):
 @operator('$let')
 def let(template, context):
     variables = renderValue(template['$let'], context)
+    
+    _let_re = re.compile(r'[a-zA-Z_][a-zA-Z0-9_]*$')
+    
+    if type(variables) != list and variables != None:
+        if not all(_let_re.match(key) for key in variables.keys()):
+            raise TemplateError('top level keys of context must follow/[a-zA-Z_][a-zA-Z0-9_]*/')
+    
     if not isinstance(variables, dict):
         raise TemplateError("$let value must evaluate to an object")
     subcontext = context.copy()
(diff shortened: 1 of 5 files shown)


jsone/render.py
```diff
@@ -124,6 +125,11 @@ def jsonConstruct(template, context):
 @operator('$let')
 def let(template, context):
     variables = renderValue(template['$let'], context)
+    
+    if type(variables) != list and variables != None:
```
djmitche: You could omit this type check if you moved this section  below the `isinstance` conditional below.

src/index.js
```diff
@@ -104,7 +104,15 @@ operators.$let = (template, context) => {
...
 
   var context_copy = Object.assign(context, variables);
-
+  
+  let eachKey = Object.keys(variables);
+ 
+  if (eachKey !== null || !isArray(eachKey)) {
+    let match = /^each\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(eachKey);
+    if (!match) {
+      throw new TemplateError('top level keys of $let must follow/[a-zA-Z_][a-zA-Z0-9_]*/');
```
djmitche: add a space before `/`

jsone/render.py
```diff
@@ -124,6 +125,11 @@ def jsonConstruct(template, context):
...
+        if not all(_let_re.match(key) for key in variables.keys()):
+            raise TemplateError('top level keys of $let must follow/[a-zA-Z_][a-zA-Z0-9_]*/')
```
djmitche: add a space before `/`

specification.yml
```diff
@@ -4788,3 +4788,25 @@ title: 'Nested error has appropriate message'
...
+
+################################################################################
+---
+section:  $let
+---
+title: 'checking name begin with number'
+context: {}
+template: {$let: {ab: 100, bb: 200, 2b: 5},
+           in: [{$eval: "ab+bb"}, {$eval: "ab-bb"}, {$eval: "ab*bb"}]}
+error: true
```
djmitche:
```yaml
error: 'top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/'
```
(same for the others)

This will check that both JS and Python give the same error message.

src/index.js
```diff
@@ -104,7 +104,15 @@ operators.$let = (template, context) => {
   let variables = render(template['$let'], context);
 
   var context_copy = Object.assign(context, variables);
-
+  
+  let eachKey = Object.keys(variables);
+ 
+  if (eachKey !== null || !isArray(eachKey)) {
+    let match = /^each\(([a-zA-Z_][a-zA-Z0-9_]*)\)$/.exec(eachKey);
```
djmitche: why is "each" in this regular expression?

src/index.js
```diff
@@ -104,7 +104,15 @@ operators.$let = (template, context) => {
   let variables = render(template['$let'], context);
 
   var context_copy = Object.assign(context, variables);
-
+  
+  let eachKey = Object.keys(variables);
```
djmitche: let's call this `keys` or `variableNames`.

src/index.js
```diff
@@ -104,7 +104,15 @@ operators.$let = (template, context) => {
...
+ 
+  if (eachKey !== null || !isArray(eachKey)) {
```
djmitche: similarly to Python, you can skip this check if you move the new clause to after the `isObject` check below.

src/index.js
djmitche: Also, `eachKey` would be an array here (`Object.keys` returns an array), so not something to which a regular expression can be applied.

djmitche:
Oh, and if you rebase onto master before your next push, the Taskcluster tests will run.

alexandrasp:
@djmitche I tried a rebase onto master but it did not work exactly how it should, so..I sent my code to let branch and merged it afterwards in my master.

djmitche:
OK, no worries.  I'll run this locally now to check.

djmitche:
Looks good!

Issue #154, opened 2017-07-13
Consider interpolating `${null}` as empty string


jonasfj:
Currently we don't allow interpolation of objects and arrays...

these seems to work:
* `${"string"}` => `"string"`
* `${55.5}` => `"55.5"`
* `${null}` => `"null"`

I propose that we change it so that: `${null}` => `""`

It's weird that `null` becomes `"null"` when objects and arrays aren't allowed.
It would seems useful to allow `null` to be used as empty string, since that just means it disappears.

Example: `"${firstName} ${middleName} ${lastName}"` if `middleName` is `null` it'll still work... as is it just becomes the `"James null Bond"`, instead of `"James Bond"`

djmitche:
This makes sense to me.  We should add a bit to the README too.

alisha17:
Can I take up this issue to work upon?

jonasfj:
@alisha17, absolutely...

Btw, we hang out on IRC in #taskcluster and #tc-contributors on irc.mozilla.org if you have any questions..

alisha17:
@jonasfj  Could you guide me a bit about which JS file does this code relate to? Maybe interpreter.js?

djmitche:
You probably want to start [here](https://github.com/taskcluster/json-e/blob/f63ffdcb269be0fbc2101ecde6facc011ae54c8b/src/index.js#L16-L45).

alisha17:
Done! Regarding the test : [link](https://github.com/taskcluster/json-e/blob/master/test/specification_test.js#L16), where can I find this SPEC_FILE?
@djmitche

djmitche:
In Node, `__dirname__` is the name of the directory containing the current source file (sort of like `__file__` in Python).  So

```
const SPEC_FILE = path.join(__dirname, '../specification.yml');
```

means that it's the file `specification.yml` in the parent directory to that containing `specification_test.js`.

djmitche:
Fixed in #178.  @alisha17 by the way if you write something like "Fixes #154" in your pull request description, then Github will automatically close that issue when the PR is merged.

alisha17:
@djmitche  Ah! Wasn't aware about it. Will take care now.

PR #178, opened 2017-09-20
Interpolating `${null}` as empty string

alisha17:
Fix to issue #154

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -27,9 +27,9 @@ let interpolate = (string, context) => {
         throw new TemplateError(`interpolation of '${input}' produced an array or object`);
       }
 
-      // toString renders null as an empty string, which is not what we want
+      // if it is null, result should just be appended with empty string
       if (v.result === null) {
-        result += 'null';
+        result += '';
       } else {
         result += v.result.toString();
       }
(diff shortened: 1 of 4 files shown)


alisha17:
@djmitche  What needs to be changed? I can't see the requested changes/comments.

djmitche:
Oops, I must not have hit "enter" on that.. sorry!

src/index.js
```diff
@@ -27,9 +27,9 @@ let interpolate = (string, context) => {
         throw new TemplateError(`interpolation of '${input}' produced an array or object`);
       }
 
-      // toString renders null as an empty string, which is not what we want
+      // if it is null, result should just be appended with empty string
       if (v.result === null) {
-        result += 'null';
+        result += '';
```
djmitche: `null.toString()` is an empty string, so this whole conditional can be compressed to just `result += v.result.toString();`

alisha17: But null.toString() might not be an empty string always and in all browsers.

djmitche: Indeed:
```
> null.toString()
TypeError: Cannot read property 'toString' of null
```
so I was 100% wrong :)

djmitche:
Sorry about that, and thanks for correcting me!

alisha17:
@djmitche  That's okay! Thanks :)

PR #179, opened 2017-09-22
Disallow {$keyword} for undefined keywords

### LINKED ISSUE #151: Reserve `$<keyword>` for future constructs...
Let's just require that things are always written with `{$$<whatever>: ...}` to get `{$<whatever>: ...}`.

That way we can safely introduce new $-keywords without breaking existing templates.

**Right now** the `{$$<keyword>: ...}` to `{$keyword: ...}` escaping only works for reserved keywords.

---

Notably, I could imagine a $zip, $unzip, $pairs, $unique, $reduce, {$call: ..., with: ...}, {$transform: '...', ...}, and others from lodash.. Arguably we should think of a way these constructs will be useful before we introduce them.

djmitche:
Fixes #151.

Hopefully I got this the right way 'round this time?

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -319,7 +319,7 @@ result:   {$sort: [3, 2, 1]}
 title:    reserved $-keywords
 context:  {}
 template: {$nothing: 1}
-error:    true
+error:    'TemplateError: $<identifier> is reserved; ues $$<identifier>'
 ---
 title:    escape non-json-e operator
 context:  {}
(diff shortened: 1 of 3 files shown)


specification.yml
```diff
@@ -316,15 +316,15 @@ context:  {}
 template: {$$sort: [3, 2, 1]}
 result:   {$sort: [3, 2, 1]}
 ---
-title:    interpolate non-json-e operator
+title:    reserved $-keywords
 context:  {}
 template: {$nothing: 1}
-result:   {$nothing: 1}
+error:    'TemplateError: $<identifier> is researved; ues $$<identifier>'
```
jonasfj: `researved` -> `reserved`

PR #182, opened 2017-09-23
Refactor the demo site into several tabs

### LINKED ISSUE #107: Split demo site into three pages
As described in https://github.com/taskcluster/json-e/pull/105#pullrequestreview-45264466

> Maybe split this into three pages (entrypoints): home, docs, and playground. The home page would have the header that's baked into app.js at the moment; the docs would be a faithful rendition of the full README (without special-casing certain headings), and the playground would be a cleaner view. It's a bit confusing right now that the first of 100 template/context/result rows is a playground, while the rest start with interesting content.

djmitche:
This handles navigation using hashes.  One neat advantage is that it keeps the formatting from the YAML, even in the playground.

Fixes #107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,9 +211,10 @@ result:   [1, 2, 3]
 ### `$fromNow`
 
 The `$fromNow` operator is a shorthand for the built-in function `fromNow`. It
-creates a JSON (ISO 8601) datestamp for a time relative to the current time or,
-if `from` is given, from that time.  The offset is specified by a sequence of
-number/unit pairs in a string. For example:
+creates a JSON (ISO 8601) datestamp for a time relative to the current time
+(see the `now` builtin, below) or, if `from` is given, relative to that time.
+The offset is specified by a sequence of number/unit pairs in a string. For
+example:
 
 ```yaml
 template: {$fromNow: '2 days 1 hour'}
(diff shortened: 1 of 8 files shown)


imbstack:
Nits:

1. Let's give a `pointer` to the mouse on hover for the tabs at the top.
2. We can pretty up the sidebar a bit. Don't put too much work in though!
3. The landing page feels a bit empty on my screen now? Maybe if we had a logo/image to go there or... well who really cares I guess
4. I kinda liked the version being in a little blue label vs how it is just plainly written alongside json-e in the title now. If you hated the old way though, I'm happy either way. 
5. When I open examples in the playground, it seems to keep the formatting including some whitespace before the inputs? Probably not worth fussing over.

r+ This is much nicer!

djmitche:
> 1. Let's give a `pointer` to the mouse on hover for the tabs at the top.
> 2. We can pretty up the sidebar a bit. Don't put too much work in though!

Will do

> 3. The landing page feels a bit empty on my screen now? Maybe if we had a logo/image to go there or... well who really cares I guess

It seems like we don't want to highlight the link of JSON-e and Taskcluster.. and I don't know what other logo to put there.  So I'm with you..

> 4. I kinda liked the version being in a little blue label vs how it is just plainly written alongside json-e in the title now. If you hated the old way though, I'm happy either way. 

I liked it, but rebass doesn't support it anymore.

> 5. When I open examples in the playground, it seems to keep the formatting including some whitespace before the inputs? Probably not worth fussing over.

Hm, maybe dedent doesn't strip single-line strings.  I'll look into it.

djmitche:
On further investigation: Badge is still supported, but it's not the nice rounded pill anymore.  The rest are fixed.

PR #183, opened 2017-09-25
checking integrity of $let for positive test and fix JS implementation for this check

alexandrasp:
@djmitche fixed! I found a new problem in JS implementation for $let check, to do the same as the python implementation I used a lambda in JS.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -104,17 +104,14 @@ operators.$let = (template, context) => {
   let variables = render(template['$let'], context);
 
   var context_copy = Object.assign(context, variables);
-  
-  let eachKey = Object.keys(variables);
- 
-  if (eachKey !== null || !isArray(eachKey)) {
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -108,7 +108,7 @@ operators.$let = (template, context) => {
   if (!isObject(variables)) {
     throw new TemplateError('$let value must evaluate to an object');
   } else {
-    let match = /^[a-zA-Z_][a-zA-Z0-9_]*$/.exec(Object.keys(variables));
+    let match = Object.keys(variables).every((variableNames) => /^[a-zA-Z_][a-zA-Z0-9_]*$/.exec(variableNames));
```
djmitche: This looks correct, but `variableNames` is a plural noun, while it only contains a singular variable name, so please change the name to something like `variableName`.

djmitche:
Nice work, thanks for your patience!

Issue #180, opened 2017-09-22
disallow str(array)


djmitche:
The behavior of `str` on an array is not currently defined.  We should either define it and test it thoroughly, or disallow it.  I'd lean toward disallowing it.  (From #155)

prachi1210:
Hi, 
I'd like to work on this issue. From https://github.com/taskcluster/json-e/pull/155 I understand that we want to disallow str(array) from just $eval, is that right?

djmitche:
Also in string interpolation -- for example

```yaml
template: "entries are ${str(entries)}"
context:
  entries: [1, 1, 2, 3, 5, 8]
```

prachi1210:
I think arrays are already disallowed in string interpolation here 
```
title:    can't interpolate arrays
context:  {key: [1,2,3]}
template: {message: 'hello ${key}'}
error:    "TemplateError at template.message: interpolation of 'key' produced an array or object"
```

PR #184, opened 2017-09-25
Disallow str(array) from $eval

prachi1210:
Closes https://github.com/taskcluster/json-e/issues/180

### DIFF
--- a/jsone/builtins.py
+++ b/jsone/builtins.py
@@ -14,7 +14,7 @@ def build(context):
     def builtin(name, variadic=None, argument_tests=None, minArgs=None):
         def wrap(fn):
             def bad(reason=None):
-                raise BuiltinError((reason or 'invalid arguments to {}').format(name))
+                raise BuiltinError((reason or 'invalid arguments to builtin: {}').format(name))
             if variadic:
                 def invoke(*args):
                     if minArgs:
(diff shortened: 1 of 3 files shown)


src/builtins.js
```diff
@@ -85,10 +85,15 @@ module.exports = (context) => {
   define('str', builtins, {
     argumentTests: ['string|number|boolean|array|null'],
```
djmitche: I think just removing `array` here would be sufficient

specification.yml
```diff
@@ -1510,15 +1510,15 @@ context: {}
...
+title: str([1, 2, 3])
 context: {}
 template: {$eval: 'str([1, 2, 3])'}
-result: '1,2,3'
+error: true
+---
+title: str(key) with array
+context: {key: [1, 2, 3]}
+template: {$eval: 'str(key)'}
+error: true
```
djmitche: Please include the actual error message here -- that will check that both the JS and Python implementations give the same message.

jsone/builtins.py
```diff
@@ -78,7 +81,7 @@ def lowercase(v):
         return v.upper()
 
     builtin('len', argument_tests=[is_string_or_array])(len)
-    builtin('str', argument_tests=[anything])(to_str)
+    builtin('str', argument_tests=[anything_except_array])(to_str)
```
djmitche: This will allow an object (`{..}`), but the JS implementation does not (the JS implementation is correct).

Can you add an extra test case for `str({a: 10})` and adjust the implementations to pass?

prachi1210:
@djmitche done!

jsone/builtins.py
prachi1210: `return isinstance(v, (string, int, float, bool)) or v is None` only allows string, number, bool or null

djmitche: Oh, I see, I only looked at the name of the function.

djmitche:
Awesome, thanks!

Issue #181, opened 2017-09-22
`now` constant appears to be a Date, not a string


djmitche:
```yaml
template: now
context: {}
result: 2017-01-19T16:27:20.974Z
```
yet
```yaml
template: {$fromNow: '0 seconds'}
context: {}
result: '2017-01-19T16:27:20.974Z'
```

This should show up as a string value.

lauryndbrown:
Hi, I'd like to work on this issue. But I'm noticing that now is simply giving me the word "now" rather than a date time stamp.

`template = "now"
context = {}
print(jsone.render(template,context))`

When trying it on the docs site (https://taskcluster.github.io/json-e/), I also simply get "now". Is there something special necessary when using the now builtin?

djmitche:
To evaluate a variable, you need to use eval.  My example is wrong.  It should be

```yaml
template: {$eval: now}
context: {}
result: 2017-01-19T16:27:20.974Z
```
sorry about that!

djmitche:
@lauryndbrown as a quick github tip, if you put "Fixes #181" in the description of the PR, then Github will automatically close that issue when the PR is merged.

PR #185, opened 2017-09-25
`now` constant appears to be a Date, not a string #181

lauryndbrown:
Fixing #181. 

After looking at the code, I could see that now was implemented by simply creating a new DateTime object. Instead, I call the $fromNow method already implemented to not create redundant code. I added two simple tests to ensure it worked properly. Feedback is welcomed.

### DIFF
--- a/jsone/__init__.py
+++ b/jsone/__init__.py
@@ -1,9 +1,8 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
-import datetime
 import re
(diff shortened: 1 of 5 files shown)


lauryndbrown:
@djmitche First pull request. Please let me know how I can improve when you can.

djmitche:
Nice work, @lauryndbrown :)

Get in touch to talk about next steps for Outreachy!

Issue #176, opened 2017-09-18
Prohibit extra keys in operations


djmitche:
```
$if: 'x == 10'
then:
foo: bar
else:
  bing: baz
```

Is subtly wrong. that `foo` property in the top-level object will be ignored, with no error.

djmitche:
So the idea here is that for an operator (`{$.., ..}`), *only* the defined properties should be allowed.  For `$merge`, for example, no other properties are allowed.  For `$if`, `then` and `else` are allowed (but not `foo`!)

prachi1210:
I'd like to work on this issue. We want to disallow undefined properties corresponding to the following operators:
- `$if-then-else`
- `$flatten`
- `$flattenDeep`
- `$merge`
Are there any that I missed?

djmitche:
All of the operators, please :)

prachi1210:
Just a clarification, disallowing undefined properties would require me to flag them as errors in specifications, right?

PR #186, opened 2017-09-28
Prohibit undefined properties for $merge

prachi1210:
See https://github.com/taskcluster/json-e/issues/176

@djmitche I have only disallowed undefined properties for `$merge` in this pull request as I was unsure of the implementation. If this is okay, I'll do it for the rest of the operators.

### DIFF
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -57,6 +57,14 @@ def interpolate(string, context):
 
     return ''.join(result)
 
+def checkUndefinedProperties(operator, template, allowed):
+    unknownKeys = ""
+    for key in sorted(template):
+        if key not in allowed and key != operator:
+            unknownKeys += " " + key
+    if unknownKeys:
+        raise TemplateError(operator + " has undefined properties:" + unknownKeys)
+
 
 @operator('$eval')
 def eval(template, context):
@@ -181,6 +189,7 @@ def map(template, context):
 
 @operator('$merge')
 def merge(template, context):
+    checkUndefinedProperties('$merge', template, [])
     value = renderValue(template['$merge'], context)
     if not isinstance(value, list) or not all(isinstance(e, dict) for e in value):
         raise TemplateError("$merge value must evaluate to an array of objects")
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,18 @@ var {
 var addBuiltins = require('./builtins');
 var {JSONTemplateError, TemplateError} = require('./error');
 
+function checkUndefinedProperties(operator, template, allowed) {
+  var unknownKeys = '';
+  for (var key of Object.keys(template).sort()) {
+    if (!allowed.includes(key) && key != operator) {
+      unknownKeys += ' ' + key;
+    }
+  }
+  if (unknownKeys) {
+    throw new TemplateError(operator + ' has undefined properties:' + unknownKeys);
+  }
+};
+
 let flattenDeep = (a) => {
   return Array.isArray(a) ? [].concat(...a.map(flattenDeep)) : a;
 };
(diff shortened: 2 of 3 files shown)


jsone/render.py
```diff
@@ -181,6 +181,8 @@ def gen():
 
 @operator('$merge')
 def merge(template, context):
+    if(len(template)>1):
```
djmitche: a little more whitespace would be good here (I thought we had set up something to check style automatically, but apparently not):
```python
if len(template) > 1:
```

jsone/render.py
```diff
@@ -181,6 +181,8 @@ def gen():
...
+        raise TemplateError("$merge with undefined properties")
```
djmitche: Let's be a little more advanced and give a message naming the extra properties: `$merge operator has unexpected properties: 'foo'`. If you sort the list of unexpected properties, that will make the message always match.

prachi1210:
Thanks for the review! 
What about 
`$merge:  foo : [{a: 1}, {b: 2, c: 3}, {d: 4}]`
Should this automatically prune the nested keys for $merge and display the output corresponding to the operator or should this be invalid too?

djmitche:
Putting braces on that to be explicit:
```js
{$merge: {foo : [{a: 1}, {b: 2, c: 3}, {d: 4}]}}
```
in this case, the value given to `$merge` is not an array, so it [gives an error](https://taskcluster.github.io/json-e/#Playground/%7B%7D&%7B%24merge%3A%20%7Bfoo%20%3A%20%5B%7Ba%3A%201%7D%2C%20%7Bb%3A%202%2C%20c%3A%203%7D%2C%20%7Bd%3A%204%7D%5D%7D%7D)

prachi1210:
@djmitche I'll be using the function `checkUndefinedProperties` for the rest of the operators too.

jsone/render.py
```diff
@@ -57,6 +57,14 @@ def interpolate(string, context):
 
     return ''.join(result)
 
+def checkUndefinedProperties(operator, template):
+    properties = ""
+    if (operator == '$merge'):
```
djmitche: Remember Python doesn't require these enclosing parentheses :)

prachi1210: Oh yes, I'll correct this.

prachi1210:
@djmitche Should I create a separate PR for the rest of the operators? This has now been simplified to just passing the allowed properties to a function and can be bundled in a single commit.

prachi1210:
My mistake, I didn't read your comment properly. Updated, now!

jsone/render.py
```diff
@@ -57,6 +57,14 @@ def interpolate(string, context):
 
     return ''.join(result)
 
+def checkUndefinedProperties(operator, template, allowed):
+    properties = ""
+    for key in sorted(template):
+        if key not in allowed and key!=operator:
```
djmitche: add spaces around `!=`

jsone/render.py
```diff
@@ -57,6 +57,14 @@ def interpolate(string, context):
...
+            properties += " "+ key
```
djmitche: add space before `+`

src/index.js
```diff
@@ -9,6 +9,18 @@ var {
 var addBuiltins = require('./builtins');
 var {JSONTemplateError, TemplateError} = require('./error');
 
+function checkUndefinedProperties(operator, template, allowed) {
+  var properties = '';
+  for (var key of Object.keys(template).sort()) {
+    if (!allowed.includes(key) && key!=operator) {
```
djmitche: add spaces around `!=`

src/index.js
```diff
@@ -9,6 +9,18 @@ var {
...
+      properties += ' '+key;
```
djmitche: add spaces around `+`

prachi1210:
@djmitche done!

PR #188, opened 2017-09-29
Prohibit extra keys for operators

prachi1210:
Operators included
``` 
$mergeDeep 
$flatten 
$flattenDeep 
$let 
$if 
$reverse 
$json 
$fromNow
```
@djmitche I am unsure of the representation for `$map` which takes `each(var)` and `$sort` which takes `by(var)`. Should I use regex to represent `var` while passing allowed keys? 

Fixes https://github.com/taskcluster/json-e/issues/176

### DIFF
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -73,6 +73,7 @@ def eval(template, context):
 
 @operator('$flatten')
 def flatten(template, context):
+    checkUndefinedProperties('$flatten', template, [])
     value = renderValue(template['$flatten'], context)
     if not isinstance(value, list):
         raise TemplateError('$flatten value must evaluate to an array')
@@ -89,6 +90,7 @@ def flatten(template, context):
 
 @operator('$flattenDeep')
 def flattenDeep(template, context):
+    checkUndefinedProperties('$flattenDeep', template, [])
     value = renderValue(template['$flattenDeep'], context)
     if not isinstance(value, list):
         raise TemplateError('$flattenDeep value must evaluate to an array')
@@ -106,6 +108,7 @@ def flattenDeep(template, context):
 
 @operator('$fromNow')
 def fromNow(template, context):
+    checkUndefinedProperties('$fromNow', template, ['from'])
     offset = renderValue(template['$fromNow'], context)
     reference = renderValue(template['from'], context) if 'from' in template else context.get('now')
 
@@ -116,6 +119,7 @@ def fromNow(template, context):
 
 @operator('$if')
 def ifConstruct(template, context):
+    checkUndefinedProperties('$if', template, ['then', 'else'])
     condition = evaluateExpression(template['$if'], context)
     try:
         if condition:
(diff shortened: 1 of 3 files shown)


djmitche:
I think that sounds like a good plan.  This PR looks good as-is.  I can merge it now, or wait until $map and $sort are ready.

prachi1210:
@djmitche I couldn't get regex to match without looping through the keys, so I settled for `startswith`. If this logic is too convulated, I can change it. I'll duplicate it for $map if it is okay

jsone/render.py
```diff
@@ -57,11 +57,12 @@ def interpolate(string, context):
 
     return ''.join(result)
 
-def checkUndefinedProperties(operator, template, allowed):
+def checkUndefinedProperties(operator, template, allowed, variableProperty=''):
```
djmitche: Rather than define a different parameter for this, what if each element in `allowed` could be either a string (expecting an exact match) or a regular expression object?  So something like:
```python
checkUndefinedProperties('$sort', template, [re.compile(r'by\([a-zA-Z_][a-zA-Z0-9_]*\)')])
```

prachi1210:
Thank you for the detailed review! 
I was avoiding this particular approach of using regex as I would have to loop twice now 
-  `for key in sorted(template):` 
-  ` if any(re.match(valid, key)) for valid in allowed`
unlike the current approach where it is simple a conditional check. I understand that it may not be the right way to go especially if we look to add operators which have more of properties with variables. 

Also, what do you think about combining all the allowed keys in the form of a single regex in the function itself, something on the lines of `COMBINED_RE = operator + "|".join(allowed)+"#"` This would help me avoid the two loops as stated above.
 
If I understood something incorrectly, please let me know.

djmitche:
I like the idea of constructing a single regular expression!

prachi1210:
@djmitche done!

src/index.js
```diff
@@ -11,8 +11,9 @@ var {JSONTemplateError, TemplateError} = require('./error');
 
 function checkUndefinedProperties(operator, template, allowed) {
   var unknownKeys = '';
+  var combined = new RegExp(allowed.join('|') + '$');
   for (var key of Object.keys(template).sort()) {
-    if (!allowed.includes(key) && key != operator) {
+    if (key != operator && (!combined.test(key) || combined.test('$'))) {
```
djmitche: What is the `combined.test('$')` for?

prachi1210: Unexpectedly, the regex for empty set of allowed properties: `$` was matching with all strings.

djmitche: Ah -- my suggestion to include the operator itself (`$merge` etc.) in the list of allowed properties would help that.  The list would never be empty!

Do you want to make that change?

With that fixed, we can get this merged :)

prachi1210: I'll just do that, sure!

PR #189, opened 2017-10-02
First part of enforce error messages

alexandrasp:


jsone/builtins.js
```diff
@@ -0,0 +1,138 @@
+var {BuiltinError} = require('./error');
```
djmitche: Did this flie get added accidentally?

jsone/interpreter.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, context):
 
     def parse(self, expression):
         if not isinstance(expression, string):
-            raise TemplateError('expression to be evaluated must be a string')
+            raise TemplateError('expression to be evaluated must be a string, an object, a number, a boolean or null')
```
djmitche: This doesn't look right....

jsone/render.py
```diff
@@ -243,7 +243,7 @@ def reverse(template, context):
     checkUndefinedProperties(template, ['\$reverse'])
     value = renderValue(template['$reverse'], context)
     if not isinstance(value, list):
-        raise TemplateError("$reverse value must evaluate to an array")
+        raise TemplateError("$reverse value must evaluate to an array of objects")
```
djmitche: This seems incorrect, too: `{$reverse: [1, 2, 3]}` is valid, but the numbers aren't objects.

specification.yml
```diff
@@ -928,81 +928,81 @@ title:    cannot sort objects without by
...
 template: {$sort: [1, 'a'], 'by(x)': 'x'}
-error: true
+error: 'TemplateError: $sort requires all sorted values have the same type or structure, not null and not boolean'
 ---
 title:    sort requires an array (string)
 context:  {}
 template:
   $sort:  "some string"
-error: true  # $sort must be given an array
+error: 'TemplateError: $sort value must evaluate to an array'  # $sort must be given an array
```
djmitche: you can remove the comment on this line and the next few copies of it

djmitche:
@alexandrasp do you want to finish this up?  It's at risk of conflicting with other patches to the error messages..

alexandrasp:
yes @djmitche, I want finish.

jsone/builtins.js
alexandrasp: yes, probably

djmitche:
Is this ready for another look?  I see that `jsone/builtins.js` still exists..

djmitche:
Also, you might want to rebase onto master..

djmitche:
Looks like this is failing some tests -- https://public-artifacts.taskcluster.net/EYPzPND_SgOqi2JLyGYMnA/0/public/logs/live_backing.log

PR #190, opened 2017-10-02
 removing unnecessary if condition in $reverse op of JS code 

alexandrasp:


src/index.js
```diff
@@ -238,10 +238,6 @@ operators.$reverse = (template, context) => {
   if (!isArray(value) && !isArray(template['$reverse'])) {
```
djmitche: Oh!  I didn't even notice the `if` you removed when I first looked at this code.  I think the `!isArray(template['$reverse'])` here is also redundant.  Is there a situation where `isArray(value)` is false but `isArray(template['$reverse'])` is true?

alexandrasp: Yes @djmitche, it makes sense, specially if we check the code...
`let value = render(template['$reverse'], context);`
Calling render that don't change an array to other type. Can it be conclusive? 
I checked the test cases with `$reverse` template...and again always `false && false` or `true&&true`.

alexandrasp: I think the code is ready to be merged ;)

djmitche: Right, render of an array is always an array.

djmitche: Let's get rid of the second part of this conditional too, then merge :)

alexandrasp: pushed :)

djmitche:
Thanks, sorry for the back-and-forth :)

alexandrasp:
no problem @djmitche ;)

Issue #161, opened 2017-07-14
proposal `$switch`


jonasfj:
```yaml
---
title: $switch on string
context: {key: "kindA"}
template:
  $switch: "key"
  kindA: {$eval: '4 - 1'}
  kindB: {$eval: '3 + 4'}
result: 3
---
title: $switch on type
context: {key: 42}
template:
  $switch: "type(key)"  # works when bstacks type/typeof function is introduced
  string: {$eval: '4 - 1'}
  number: {$eval: '3 + 4'}
result: 7
---
title: $switch on number
context: {key: 42}
template:
  $switch: "key"
  kindA: {$eval: '4 - 1'}
  "42": {$eval: '3 + 4'}
result: 7
---
title: $switch without match
context: {key: 42}
template:
  $switch: "key"
  "43": 1
result: null
---
title: $switch without branches
context: {key: 42}
template:
  $switch: "key"
result: null
---
title: $switch without match as deleteMarker
context: {key: 42}
template:
  required: 23
  optional:
    $switch: "key"
    "43": 1
result: {required: 23}
---
```

djmitche:
What's the use-case for this?  I'm not opposed except on the basis of "avoid unnecessary complexity".

jonasfj:
It's thought as a general alternative to if-elseif which we don't have.
But in particular I think it would be useful with a `typeof/type` function to allow type switching:
```yaml
$switch: "type(key)"  # works when bstacks type/typeof function is introduced
  string: 7
  number: {$eval: '3 + key'}
  object: {$eval: 'key.value'}
```

jonasfj:
Imagine updating artifacts, you could switch on `artifacts[i].kind` and inside a `$map` to treat folder and file artifacts differently.

djmitche:
OK, I'm sold :)

djmitche:
I'm writing `.taskcluster.yml` for Gecko, and it would be useful there, too.

djmitche:
@owlishdeveloper one of the things I don't like about the proposal here is that it stringifies the value:

```yaml
$switch: somevalue
13: "thirteen"
twelve: "twelve"
```

Then both `13` (the number) and `"13"` (the string) will produce `"thirteen"` as output.

You suggested `$match`, and that could potentially contain condition expressions in strings.  Something like
```yaml
$match:
  - "somevalue == 13": "thirteen (number)"
  - "somevalue == '13'": "thirteen (string)"
[...]

PR #191, opened 2017-10-04
WIP(Switch implementation in js)

iamneha:
#161

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -195,6 +195,18 @@ operators.$merge = (template, context) => {
   return Object.assign({}, ...value);
 };
 
+operators.$switch = (template, context) => {
+  checkUndefinedProperties(template, ['\\switch']);
+
+  let value = render(template['$switch'], context);
+
+  if (!isString(template['$switch'])) {
+    throw new TemplateError('$switch can evaluate string expressions only');
+  }
+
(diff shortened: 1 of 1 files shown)


imbstack:
This is looking good so far! I think the next step would be to add the yaml from [the issue](https://github.com/taskcluster/json-e/issues/161#issue-243104748) to `specification.yml` and then run the js tests with `yarn test`. Once the tests pass again, we can start work on the Python bit! Let me know if you have any questions, I'll be around in irc and here!

src/index.js
```diff
@@ -195,6 +195,18 @@ operators.$merge = (template, context) => {
   return Object.assign({}, ...value);
 };
 
+operators.$switch = (template, context) => {
+  checkUndefinedProperties(template, ['\\switch']);
+
+  let value = render(template['$switch'], context);
```
jonasfj: This should be evaluate as expression in the expression language, not as json-e expression... This is what `$if` does with interpreter...

src/index.js
```diff
@@ -195,6 +195,18 @@ operators.$merge = (template, context) => {
...
+
+  if (!isString(template['$switch'])) {
```
jonasfj: The value resulting from interpreting this expression...

djmitche:
We went with `$match` with a somewhat different syntax..

Issue #168, opened 2017-08-10
enforce all error messages match


djmitche:
I got started in #166, but ideally all errors from both implementations would have the same (and thoughtful) message.

alexandrasp:
working on it  :)

djmitche:
Only 352 test cases left :)

alexandrasp:
and probably some eventual bugs :)

prachi1210:
I can help out with some of the test cases @alexandrasp @djimitche. Are you working on test cases for any of the  operators currently or is everything up for grabs?

djmitche:
I don't have any work in progress, so just please coordinate with @alexandrasp

alexandrasp:
@prachi1210, I am working in sections: `$reverse`, `$eval`, `$sort` and `$builtins`. Feel free to help with other sections o/ 

Tip: @djmitche said to me that JS messages sometimes could not be a good reference:/

djmitche:
There's still work to do here.  To find it, look for `error: true` in `specification.yml`.  This means "there should be an error" but doesn't say what the error should be.

So, to make a PR to address this, find a section of that file containing `error: true` and change it to give a string (for example, `error: 'InterpreterError: infix: ** expects number ** number'`), then test that all languages give that exact error message.  If not, figure out what the best message is and make all languages give that message.

There's room for several people to work on different sections of the file, so as alex did above, say here what you're working on to make sure others work on a different portion.

terrameijar:
@owlishDeveloper I am interested in taking this one on as well. One question I have is how do I "test that all languages give that exact error message" ?

owlishDeveloper:
@terrameijar you must setup your development environment and send me the screenshots of the successful test runs first. Please do not work on this issue before I get your screenshots. Thank you!

terrameijar:
@owlishDeveloper I have the environment setup, and the tests pass. I sent you screenshots to your email 2 days ago.

owlishDeveloper:
@terrameijar oh you're Vuyisile! I haven't realized this was your github handle. Sure, you can work on this issue!

terrameijar:
@owlishDeveloper Thanks for assigning the issue to me. I hadn't realised there are so many test cases to go through :D This will take me a few days.

terrameijar:
Added error messages for `builtins` and `infix` errors. 261 test cases to go.

owlishDeveloper:
@terrameijar it's ok to split the work and make smaller PRs!

[...]

PR #194, opened 2017-10-16
Make error messages match for infix, unary op, identifiers

prachi1210:
Partial work for https://github.com/taskcluster/json-e/issues/168

### DIFF
--- a/jsone/interpreter.py
+++ b/jsone/interpreter.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 from .prattparser import PrattParser, infix, prefix
-from .shared import TemplateError, string
+from .shared import TemplateError, InterpreterError, string
 import operator
 import json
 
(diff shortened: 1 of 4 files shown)


jsone/interpreter.py
```diff
@@ -22,7 +22,8 @@
 
 
 def expectationError(operator, expected):
```
djmitche: this should probably be renamed to `infixExpectationError` since it's specific to infix operators now.

prachi1210: This function is also called while throwing expectation error for unary operators. Should I still rename it?

djmitche: https://github.com/prachi1210/json-e/blob/1d29d55bffd4967ba61b2e6472abbfc080586ba6/jsone/interpreter.py#L80-L92 is raising InterpreterErrors directly.

djmitche: @prachi1210 do you want to finish this up?

prachi1210:
Yes. @djmitche I was taken sick the past week. Will get to it asap!

djmitche:
@prachi1210 I added that rename and I'll get this merged now once it passes tests.

PR #195, opened 2017-10-16
Make error messages match compound literals

prachi1210:
Partial work for https://github.com/taskcluster/json-e/issues/168


djmitche:
If you switch to the `error-match-compound-literals` branch and use `git rebase -i`, you can drop the infix and unary commits and only have the compound literal commit left.  Then `git push -f` and you can open a new PR (or we can reopen this one).

PR #198, opened 2017-10-27
Operators || and && evaluates right too soon

jonasfj:
Test cases for this is needed, we can check by adding an interpreting error on the right hand side..

THIS PR is really a bug report, with a partial fix, test-cases are needed. Please don't merge without tests...


djmitche:
Fixed in #205.

PR #205, opened 2017-11-27
Precedence of `in` relative to `||`

### LINKED ISSUE #201: Precedence of `ìn` is wrong relative to `||`
Example: `{$eval: "'a' in ['b'] || true"}` does not work...
while: `{$eval: "('a' in ['b']) || true"}` does work.

We probably need to review the precedence of the `in` operator.

ydidwania:
Fixes #201 
Added the three test cases to `specification.yml`  and changed to precedence of `in` . It now looks like this 
```
 precedence: [
    ['||'],
    ['&&'],
    ['in'],
    ['==', '!='],
    ['>=', '<=', '<', '>'],
    ... 
 ]
```
All tests pass successfully!

### DIFF
--- a/jsone/interpreter.py
+++ b/jsone/interpreter.py
@@ -44,9 +44,9 @@ class ExpressionEvaluator(PrattParser):
         'null', 'number', 'identifier', 'string',
     ]
     precedence = [
-        ['in'],
         ['||'],
         ['&&'],
+        ['in'],
         ['==', '!='],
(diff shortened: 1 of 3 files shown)


djmitche:
Looks like you need to fix the precedence for Python, too, and this should be ready to go!

ydidwania:
Oops! Forgot that :)

djmitche:
👍 thanks!

PR #206, opened 2017-11-29
Ensure date has microseconds

### LINKED ISSUE #200: Invalid date-time can be generated, intermittently
STR:
 - be on halloween day off
 - submit taskcluster tasks

Expected:
 Works.

Actual:
 Fails with "InputValidationError"

This is the error:
```
{
  "code": "InputValidationError",
   ...
...
      "created": "2017-11-01T22:00:09",
...
}
```

As you can see, `created`, `deadline` and `expires` are all invalid because they lack the timezone info. It is generated here [https://github.com/mozilla/DeepSpeech/blob/12dee36a43fbbebcccf1a3c2cbc8946c54924f7d/taskcluster/linux-opt-base.tyml#L10-L15]:
```
created: { $fromNow: '0 sec' }
deadline: { $fromNow: '1 day' }
expires:
  $if: 'event.event == "push"'
  then: { $fromNow: '6 months' }
else: { $fromNow: '7 days' }
```

This template is being ingested by JSON-e Python module.

*[edit: removed literal newlines as they are confusing TaskWarrior]*

lissyx:
Fixes #200

### DIFF
--- a/jsone/shared.py
+++ b/jsone/shared.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 import re
 import datetime
+import six
 
 
 class DeleteMarker:
@@ -113,7 +114,13 @@ def to_str(v):
 
 def stringDate(date):
     # Convert to isoFormat
-    string = date.isoformat()
+    if six.PY3:
+        string = date.isoformat(timespec='microseconds')
+    # py2.7 does not have timespec
+    elif six.PY2:
+        string = date.isoformat()
+        if string.find('.') == -1:
+            string += '.000'
     string = datefmt_re.sub(r'\1Z', string)
     return string
 
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -1,8 +1,9 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 import math
+import datetime
 from nose.tools import eq_
-from jsone.shared import string 
+from jsone.shared import string, stringDate
 from jsone import render, JSONTemplateError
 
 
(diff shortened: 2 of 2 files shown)


jsone/shared.py
```diff
@@ -113,7 +113,13 @@ def to_str(v):
 
 def stringDate(date):
     # Convert to isoFormat
-    string = date.isoformat()
+    try:
+        string = date.isoformat(timespec='microseconds')
+    # py2.7 does not have timespec
+    except TypeError as e:
```
djmitche: Rather than incur an exception on every invocation, we could define `stringDate` differently for PY2 and PY3.  We use `six` wihch has [constants for that purpose](http://six.readthedocs.io/#package-contents).

jsone/shared.py
```diff
@@ -113,7 +113,13 @@ def to_str(v):
...
+        string = date.isoformat()
+        if string.find('.') == -1:
+            string += '.000'
```
djmitche: I think this solution is good for py27.

jsone/shared.py
lissyx: Thanks, I forgot about `six` and I was unable to find it earlier I'll use that :)

lissyx:
@djmitche Using `six.PY2`/`six.PY3` :)

jsone/shared.py
```diff
@@ -113,7 +114,13 @@ def to_str(v):
 
 def stringDate(date):
     # Convert to isoFormat
-    string = date.isoformat()
+    if six.PY3:
```
djmitche: Better!  I was thinking of conditionalizing the definition of the function on PY3:

```python
if PY3:
    def stringDate(date): ..
else:
    def stringDate(data) ..
```

Still, this is good, so I'll get it merged and released.

PR #207, opened 2017-11-30
Ensure date has microseconds

### LINKED ISSUE #200: Invalid date-time can be generated, intermittently

lissyx:
Fixes #200

### DIFF
--- a/jsone/shared.py
+++ b/jsone/shared.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 import re
 import datetime
-import six
 
 
 class DeleteMarker:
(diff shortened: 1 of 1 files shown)


lissyx:
According to https://github.com/python/cpython/commit/a2998a63c8951ab8254b96860fe64dcaa215eb1f#diff-25e2d173c84057d069b7890450714edd, it would be available since 3.5.2.

PR #209, opened 2017-12-04
Golang port

jonasfj:
The spec changes were necessary because:
 * some errors are input errors by the person constructing the context, these should cause panic in golang
 * yaml parser in golang is YAML 1.1, which means `{y: 1}`  becomes `{"true": 1}`, see: https://github.com/go-yaml/yaml/issues/214

other than that, it passes all test cases :)

We should hook up automation too...


jonasfj:
I should probably write some documentation too...

djmitche:
Merging this means any subsequent modification requires implementation in all three languages.  Is that worth it?

jonasfj:
> Merging this means any subsequent modification requires implementation in all three languages. Is that worth it?

Yes, I want to do a rust port too... :)

FYI, I would love to use json-e for configuration loading in `taskcluster-worker`, I would also like to use it to replace `typed-env-config` in our node libraries, granted that's less important.

json-e should see additional bug fixes for various poorly defined semantics and sketchy things going on.
But I don't see us implementing major features and changes going forward. For most things we want to add we should add them by:
 * Adding new `{$<identifier>: ..., ...}` constructs (I see us adding few of these)
 * Adding new built-in functions

Both of these are easy to add in all there languages, as they are just declarative entries in a dictionary.

Most functionality can already be achieved by adding custom functions to your context in places where you are using json-e.

I see lots of value in having this implemented in different languages. Given how similar the implementations are fixing any remaining operator precedence bugs and other bugs in the semantics should be easy.

djmitche:
OK, sounds good -- I just wanted to consider the impact :)

interpreter/execute.go
```diff
@@ -0,0 +1,22 @@
+package interpreter
+
+// Execute expression with given context starting from offset.
+//
+// Values of context must be calid, see IsValidContext()
```
djmitche: *valid

interpreter/util.go
```diff
@@ -0,0 +1,384 @@
...
+		panic(err)
+	}
+	return v
+}
+
+// WrapFunctionWithContext wraps a function for usage in context passed to
+// the interpreter, panics if function is not valid.
+//
+// This requires the function to take map[string]interface{} as first argument,
+// this will be the interpret context.
```
djmitche: interpreter

interpreter/util.go
```diff
@@ -0,0 +1,384 @@
...
+
+func injectedFunctionError(f interface{}, m string) error {
+	fn := runtime.FuncForPC(reflect.ValueOf(f).Pointer())
+	file, line := fn.FileLine(fn.Entry())
+	return fmt.Errorf("injected function %s:%d %s %s", file, line, fn, m)
+}
+
+// IsWrappedFunction returns true, if the value is a function wrapped by
+// WrapFunction
+func IsWrappedFunction(value interface{}) bool {
```
djmitche: I vaguely see what this wrapped-function stuff is doing, but it's definitely deeper Golang than I'm qualified to review :)

interpreter/util.go
```diff
@@ -0,0 +1,384 @@
...
+
+	// Interpret result values
+	if len(values) == 1 || values[1].Interface() == nil {
+		return values[0].Interface(), nil
+	}
+	return values[0].Interface(), values[1].Interface().(error)
+}
+
+// IsValidateData returns an error explaining why data isn't valid, or nil
+func IsValidateData(data interface{}) error {
```
djmitche: IsValidData?

jsone.go
```diff
@@ -0,0 +1,906 @@
...
+	hours += 24 * days
+	minutes += 60 * hours
+	seconds += 60 * minutes
+
+	result := reference.Add(time.Duration(seconds*neg) * time.Second)
+
+	return result.UTC().Format(timeFormat), nil
+}
+
+var builtin = map[string]interface{}{
```
djmitche: `builtins.go`

jsone.go
```diff
@@ -0,0 +1,906 @@
...
+			)
+		}
+		return fromNow(offset, ref)
+	}),
+}
+
+var eachKeyPattern = regexp.MustCompile(`^each\(([a-zA-Z_][a-zA-Z0-9_]*)\)$`)
+var byKeyPattern = regexp.MustCompile(`^by\(([a-zA-Z_][a-zA-Z0-9_]*)\)$`)
+
+var operators = map[string]operator{
```
djmitche: `operators.go`

jsone.go
```diff
@@ -0,0 +1,906 @@
...
+	`(?:\s*(\d+)\s*d(?:ays?)?)?`,
+	`(?:\s*(\d+)\s*h(?:(?:ours?)|r)?)?`,
+	`(?:\s*(\d+)\s*m(?:in(?:utes?)?)?)?`,
+	`(?:\s*(\d+)\s*s(?:ec(?:onds?)?)?)?`,
+	`\s*$`,
+}, ""))
+
+const timeFormat = "2006-01-02T15:04:05.000Z"
+
+func fromNow(s string, reference time.Time) (string, error) {
```
djmitche: `fromNow.go`

jsone_test.go
```diff
@@ -0,0 +1,118 @@
...
+		require.Error(t, err)
+	}
+}
+
+// TestSpec will load specification.yml into testCase structs and call
+// testCase.Test(t) for each test case.
+func TestSpec(t *testing.T) {
+	// NOTE: this does some ugly YAML hacks to get around type mismatches
+	// between YAML and JSON in golang, as well as lack of document support
+	// in gopkg.in/yaml.v2
```
djmitche: Another option here may be to use some Python magic to reflect `specification.yml` into `specification.json` (not checked in).

interpreter/execute.go
```diff
@@ -0,0 +1,22 @@
...
+func Execute(expression string, offset int, context map[string]interface{}) (interface{}, error) {
+	if err := IsValidContext(context); err != nil {
+		panic(err)
+	}
```
djmitche: Could this just return an error?  Most apps will build context internally, making this a code error and thus appropriate for a panic, but some might build context from user-provided input, and would want to gracefully recover from this situation.  For example, tc-hooks will take context from the payload of the `triggerHook` method.

If this does return an error, then there's no need for the `panic` property in `specification.yml`, which is handy since "panic" doesn't mean much in Python or JS.

jonasfj: > For example, tc-hooks will take context from the payload of the triggerHook method.

But the payload from `triggerHook` is valid JSON, as JSON can _always_ be assigned to variable in the context. `context = {myVariable: anyJSON}` is okay, as long as `anyJSON` is valid JSON.

If you insert functions that can't be wrapped, or invalid top-level variables names that'll cause panic. And you should never do that.

jonasfj:
> Govendor production and test requirements. It's impressive that the only non-builtin you've referenced in production code is pkg/error! Hm, does govendor work for libraries?

You're not supposed to do anything... one can do it for running tests in CI...
But the one thing we depend on here is not likely to change they are used in many projects...

jsone.go
jonasfj: maybe, I hate polluting the top-level folder more than necessary... that's why I put the interpreter in it's own package... even if it's not really reusable..

djmitche: Yeah, these could be in a sub-package too.. I don't think go's artificial linking of repo structure to module paths should force us to lump everything into one file.

Issue #212, opened 2017-12-21
$let shouldn't render it's value, only the values of keys assigned (no dynamic variables)


jonasfj:
Example:
```yml
$let:
  ${'na' + 'me'}: jonasfj
in: 'hello ${name}'
```
Renders to `'hello jonasfj'`.

I propose that we forbid the `${'na' + 'me'}` part... 

Instead we require that in `{$let: X, in: T}`, `T` is a template as is now, while we require that `X` is an object mapping from `^[a-zA-Z_][a-zA-Z0-9_]*$` to templates, and that we render these templates.

This prevents crazy things like what I did above... It also ensures that static analysis can be used to determine which global variables are accessed. A feature that is useful in certain scenarios.

----
Note: I recognize that this will require a specification change, but that:
 1) it's unlikely that anyone is using this feature yet
 2) usage of this pattern is likely an anti-pattern... as there is no way to access top-levels variables if and only if they've been declared

djmitche:
Agree :)

jonasfj:
This was fixed in #214

PR #214, opened 2017-12-22
Let static vars

jonasfj:
Fixing #212 , note this is based on #213 to avoid conflicts...

### DIFF
--- a/jsone.go
+++ b/jsone.go
@@ -418,11 +418,7 @@ var operators = map[string]operator{
 		if err := restrictProperties(template, "$let", "in"); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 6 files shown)


jsone.go
```diff
@@ -433,8 +429,12 @@ var operators = map[string]operator{
 		for k, v := range context {
 			c[k] = v
 		}
+		var err error
 		for k, v := range o {
-			c[k] = v
+			c[k], err = render(v, context)
```
djmitche: This ends up calling `render` for the context value before checking the form of the key, which is different from the other implementations.  That doesn't affect the correctness, but might cause different error messages for a template with >1 errors.

Issue #203, opened 2017-11-12
string slicing should be encoding aware, use characters not bytes for offsets


jonasfj:
We implement json-e in python, js and in some future other languages. These all encoding strings differently.

Meaning that string slicing like: `myString[2:]` have to be encoding aware. We can't just cut after 2 bytes, if the string is non-ascii then this could have different results in python and js, as python uses utf8 and js uses uft16 for internal string encoding.

The safest thing is probably to scan the string and use character offsets. We would then have to do the same in `len(myString)` as this would have to be the number of characters (regardless of the number of bytes).

djmitche:
We should probably require everything passed to json-e to be unicode, in fact -- not bytes.

Python does not use utf8 -- for unicode strings it uses unicode character points, and for bytestrings it uses bytes.  It has functions to encode and decode between the two, and utf-8 is among the options for encodings.

jonasfj:
Oh, I thought python did uft8, but Unicode is perfect...

Anyways, yeah should operate on Unicode strings.. in Python that's just converting to Unicode strings... But how do you do that in js?

Probably well have to write methods for length and string slicing that use codepoint offsets...

prachi1210:
Hi, 
Can I work on this?

jonasfj:
I see @djmitche self-assigned, but I'm sure you're welcome to chip in...

I'm guessing that fixing the `len()` function to return number of characters rather than bytes would be a good start...

Note. I made this bug slightly more complicated by adding a golang implementation :)

djmitche:
Hi Prachi!  It's good to hear from you again.  You are welcome to work on this.

Probably the best place to start is to write some tests in specification.yml that include multi-byte unicode characters, and then see how the different implementations behave.  As Jonas suggests, `len()` is a good place to start, with something like

```
title: len of multi-byte unicode character
context: {key: "☃”}
template: {$eval: 'len(key)'}
result: 1
```

I'm guessing, but I bet the Python implementation will get this right, while the JS and Golang implementations will not.  Let's see how that goes!

prachi1210:
@djmitche both the JS and Python implementations seem to get it right..
But I'll check for Golang

prachi1210:
When I try to run `go test jsone_test.go`  I am getting an `undefined:Render` error even though go get was successful. Am I doing something wrong?

djmitche:
Hm, it worked for me with go version 1.8.  What version are you using?

djmitche:
[...]

PR #217, opened 2017-12-28
Modified len to be encoding aware

prachi1210:
Fixes len for https://github.com/taskcluster/json-e/issues/203

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -1530,6 +1530,11 @@ context: {key: 'abc'}
 template: {$eval: 'len(key)'}
 result: 3
 ---
+title: len of grinning face unicode character
+context: {key: "\U0001F601"}
+template: {$eval: 'len(key)'}
+result: 1
+---
 title: len (4)
 context: {key: [1,2,3]}
 template: {$eval: 'len("abc")'}
--- a/src/builtins.js
+++ b/src/builtins.js
@@ -94,7 +94,7 @@ module.exports = (context) => {
 
   define('len', builtins, {
     argumentTests: ['string|array'],
-    invoke: obj => obj.length,
+    invoke: obj => Array.from(obj).length,
   });
 
   define('strip', builtins, {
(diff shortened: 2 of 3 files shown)


src/builtins.js
```diff
@@ -94,7 +94,7 @@ module.exports = (context) => {
 
   define('len', builtins, {
     argumentTests: ['string|array'],
-    invoke: obj => obj.length,
+    invoke: obj => Array.from(obj).length,
```
djmitche: Very nice!

specification.yml
```diff
@@ -1530,6 +1530,11 @@ context: {key: 'abc'}
 template: {$eval: 'len(key)'}
 result: 3
 ---
+title: len of grinning face unicode character
+context: {key: "\U0001F601"}
+template: {$eval: 'len(key)'}
```
jonasfj: I would suggest adding a test case with multiple special chars.... And different encoding lengths in utf8 and utf16.

src/builtins.js
jonasfj: Does this work with chars that take two utf16 points...? Also how well supported is this?

Babel won't compile this nicely... Because it used the Array prototype.

jonasfj:
Just suggestions... They can be fixed in other PRs...

src/builtins.js
prachi1210: The grinning face uses 2 utf-16 points. That's why I used that as a sample test case.

specification.yml
prachi1210: I will do that in the subsequent pull requests.

djmitche:
With node 6, which is the oldest node we support:
```
> key="😆"
'😆'
> key.length
2
> Array.from(key).length
1
```
and in the Firefox console:
```
key = "😆"
"😆"
key.length
2
Array.from(key).length
1
```

So, I don't think there's anything Babel-specific here.  It's just using a normal, built-in behavior of Array that has been around forever, and [`String.prototype[@@iterator]()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/@@iterator)'s property of being unicode-aware.

Issue #208, opened 2017-12-04
`$map` on objects is weird when `each` results on an array


jonasfj:
```yml
$map: {hello: 5, test: 9}
each(x):
 - '${x.key}'
 - {$eval: 'x.val+1'}
```
results in:
```yml
'0': test
'1': 10
```

I think it reads the arrays as objects, perhaps it's better to throw an error.

@imbstack ^ is this intended behavior?

djmitche:
I bet it doesn't do that in Python, so probably not expected..

imbstack:
Indeed

```python
jsone.render({'$map': {'hello': 5, 'test': 9}, 'each(x)': ['${x.key}', {'$eval': 'x.val+1'}]}, {})
```

results in 

```
ValueError: dictionary update sequence element #0 has length 4; 2 is required
```

So yeah, this is a bug I expect.

jonasfj:
probably we should make the python implementation throw a nicer error too :)

note, in the golang implementation I did an if-else and two different flows depending on whether I'm mapping an array or an object.. that made the code more readable too..

Biboswan:
Hi,

As per code and docs **Given an array, map returns an array, and given an object, the result is an object**.
But here in the example $map is an object but each(x) is an array so how can the arrays be merged into a final object.

What shall be the correct output?

djmitche:
I think it should be an error when `$map` is going to return an object but `each(x)` returns something that is not an object.

Biboswan:
We can throw an error or is it better to return something like [[hello,6],[test,10]] in such case.We are getting this array result normally then since $map is an object it tries to make that array into an object.

djmitche:
It's better to throw an error.

In general, languages should "error out" on bad inputs rather than do something clever.  If there is ever a consensus on what the language should do in that situation, it's easy to turn an error into some defined behavior.  It's *not* easy to change one clever behavior into another, as that breaks compatibility.

Biboswan:
Alright,as you say.

jonasfj:
> or is it better to return something like [[hello,6],[test,10]] in such case

I agree with dustin, it's better to throw an error... If in some future we want to invent a smarter logic we can do so later...

PR #219, opened 2018-01-04
Issue #208 ( $map` on objects is weird when `each` results on an array)

Biboswan:
Added a test case .
Throwing an error if each(x)  does not  compute to an object.


specification.yml
```diff
@@ -309,6 +309,11 @@ context:  {}
 template: {$$map: [2, 4, 6], each(x): {$$eval: "x + a"}}
 result:   {$map: [2, 4, 6], each(x): {$eval: "x + a"}}
 ---
+title:	  escape $map as object and each(identifier) as array
+context:  {}
+template: {$map: {"hello": 5, "test": 9},"each(x)":['${x.key}',{$eval: 'x.val+1'}]}
+error:   'TemplateError: $map is an object so each(identifier) shall also compute to an object'
```
jonasfj: we can do a better error message I'm sure :)

src/builtins.js
```diff
@@ -62,7 +62,7 @@ module.exports = (context) => {
   });
 
   ['sqrt', 'ceil', 'floor', 'abs'].forEach(name => {
-    if (Math[name] == undefined) {
+    if (Math[name] == undefined) {
```
jonasfj: ?

src/index.js
```diff
@@ -176,6 +176,10 @@ operators.$map = (template, context) => {
 
   if (object) {
     value = Object.keys(value).map(key => ({key, val: value[key]}));
+    let firstRender= render(each, Object.assign({}, context, {[x]: value[0]}));
+    if  (!isObject(firstRender)) {
+      throw new TemplateError('$map is an object so each(identifier) shall also compute to an object');
+    }
   }
```
jonasfj: Have a look at the golang implementation... this is hard to read... I propose refactoring the entire method into:
```js
if (isObject(value)) {
  ....
  return ...;
} else {
  ....
  return ...;
}
```

even if that means we have to have a little code duplication, I think this will be prettier in the long run.

src/index.js
```diff
@@ -176,6 +176,10 @@ operators.$map = (template, context) => {
 
   if (object) {
     value = Object.keys(value).map(key => ({key, val: value[key]}));
+    let firstRender= render(each, Object.assign({}, context, {[x]: value[0]}));
```
jonasfj: Also `yarn run lint` will run a linter on the code... it's similar to `yarn test`.. :)

src/builtins.js
Biboswan: I just wrote a comment earlier which I just removed now

specification.yml
Biboswan: yes.:)Invalid input..

src/index.js
Biboswan: Yeah it ran  successfully

djmitche: It should have flagged the trailing whitespace on line 65 and the missing ` ` on this line..

specification.yml
```diff
@@ -309,6 +309,11 @@ context:  {}
 template: {$$map: [2, 4, 6], each(x): {$$eval: "x + a"}}
 result:   {$map: [2, 4, 6], each(x): {$eval: "x + a"}}
 ---
+title:    escape $map as object and each(identifier) as array
```
djmitche: This test isn't for escaping (that's the `$$`).  So, good test, but should be moved to the section with all of the other $map tests.

src/index.js
```diff
@@ -176,14 +176,16 @@ operators.$map = (template, context) => {
 
   if (object) {
     value = Object.keys(value).map(key => ({key, val: value[key]}));
-  }
-
-  value = value.map(v => render(each, Object.assign({}, context, {[x]: v})))
-    .filter(v => v !== deleteMarker);
-
-  if (object) {
+    let FirstEachValue = render(each, Object.assign({}, context, {[x]: value[0]}));
```
djmitche: Variable names should be interCaps without an initial capital letter (so `firstEachValue`)

src/index.js
```diff
@@ -176,14 +176,16 @@ operators.$map = (template, context) => {
...
+    if  (!isObject(FirstEachValue)) {
+      throw new TemplateError(`$map on objects expects each(${x}) to evaluate to an object`);
+    }
+    value = value.map(v => render(each, Object.assign({}, context, {[x]: v})))
```
djmitche: Hm, this ends up rendering the first value twice -- which is almost OK but may be surprising if someone is using a custom function that's called more of then than expected.  But more importantly, it only checks the type of the first value.  Why not check them all in this loop?

Biboswan: Then shall generator function be used here like in python implementation.

djmitche: I don't think you need a generator, but `=>` is already defining a function, so it just needs an additional statement in that function.

README.md
```diff
@@ -303,18 +301,33 @@ context:  {a: 1}
...
 ```
 
+The array or object is the value of the `$map` property, and the expression to evaluate
+is given by `each(var)` where `var` is the name of the variable containing each
+element. In the case of iterating over an object, `var` will be an object with two keys:
+`key` and `val`. These keys correspond to a key in the object and its corresponding value.
+
+When given an object, map returns an object only when the expression defined by `each(var)` 
+evaluates to an object since results of each 'each(var)' are merged internally to give the 
+resulting object.Otherwise the expression becomes invalid for the $map operator.
```
djmitche: Missing ` ` between `.` and `O`

README.md
```diff
@@ -303,18 +301,33 @@ context:  {a: 1}
...
 ```
+ 
+```yaml
+template: 
+  $map: {"hello": 5, "test": 9}
+  each(x):['${x.key}',{$eval: 'x.val+1'}]
+context: {}
+error:   'TemplateError: $map on objects expects each(x) to evaluate to an object'
+```
+
```
djmitche: I don't think we need to include error cases in the README..

Biboswan: I also think so.

README.md
djmitche: @Biboswan can you push a fix for this, too, while we wait for @jonasfj's feedback?

Biboswan: Didn't get, missing what??

Biboswan: ok got it space?

README.md
```diff
@@ -303,6 +301,16 @@ context:  {a: 1}
 result:   [3, 5, 7]
 ```
 
+The array or object is the value of the `$map` property, and the expression to evaluate
+is given by `each(var)` where `var` is the name of the variable containing each
+element. In the case of iterating over an object, `var` will be an object with two keys:
+`key` and `val`. These keys correspond to a key in the object and its corresponding value.
+
+When given an object, map returns an object only when the expression defined by `each(var)` 
+evaluates to an object since results of each 'each(var)' are merged internally to give the
```
jonasfj: nit: but this can be written more precisely...

'only when' is not a good way to describe the legal constructs...

Something like:
> When `$map` is given an object, the `each(var)` property must evaluate to an object for each key/value pair `var = {key, val}` in the object given. The objects constructed by by `each(var)` will then be merged together, with later keys overwriting previous keys.

jsone/render.py
```diff
@@ -183,21 +183,22 @@ def map(template, context):
...
         for elt in value:
             subcontext[each_var] = elt
             elt = renderValue(each_template, subcontext)
             if elt is not DeleteMarker:
                 yield elt
-
     if is_obj:
+        value = [{'key': v[0], 'val': v[1]} for v in value.items()]
         v = dict()
         for e in gen():
```
jonasfj: is `gen` referencing a variable that is declared very early in this function and then changed between the definition of `gen` and the usage of `gen`? That seems unfortunate and very hard to follow.

either pass value to `gen` as a parameter, or something else... I think passing as parameter would make a lot of sense in this case...

Note: this code actually works, but if you analyze why, it's because `value = ` is declared before `gen`... it's very hard to follow.

jsone/render.py
```diff
@@ -183,21 +183,22 @@ def map(template, context):
...
-            v.update(e)
+            if not isinstance(e, dict):
+                raise TemplateError(
+                    "$map on objects expects {0} to evaluate to an object".format(each_key))
+            else:
```
jonasfj: `else` is not needed here...
If we `raise` an exception then that's is going to break execution and return from the function...

src/index.js
```diff
@@ -173,17 +173,22 @@ operators.$map = (template, context) => {
...
-  value = value.map(v => render(each, Object.assign({}, context, {[x]: v})))
-    .filter(v => v !== deleteMarker);
-
-  if (object) {
+    let eachValue;
+    value = value.map(v => {
+      eachValue = render(each, Object.assign({}, context, {[x]: v}));
+      if  (!isObject(eachValue)) {
+        throw new TemplateError(`$map on objects expects each(${x}) to evaluate to an object`);
+      } else {
```
jonasfj: again the `else` is not needed here... if we throw an exception is going to break execution of the function

src/index.js
```diff
@@ -173,17 +173,22 @@ operators.$map = (template, context) => {
...
+        return eachValue;
+      }
+    }).filter(v => v !== deleteMarker);
     return value.reduce((a, o) => Object.assign(a, o), {});
```
jonasfj: I think we can do: `return Object.assign({}, ...value);`, granted I might be wrong...

src/index.js
```diff
@@ -173,17 +173,22 @@ operators.$map = (template, context) => {
...
   } else {
+    value = value.map(v => render(each, Object.assign({}, context, {[x]: v})))
+      .filter(v => v !== deleteMarker);
     return value;
```
jonasfj: Just do a:
```js
return value.map(
  v => render(....)
).filter(....);
```
or something like that...

Writing `value = value` followed by `return value` doesn't make it easier to read...

Biboswan:
@ jonasfj  now its fine I think

djmitche:
Awesome!

PR #221, opened 2018-01-06
bug 1328719 - sort $json output.

escapewindow:
Since we need to be able to deterministically rebuild task definitions
via json-e, we need the output to be deterministic. This means sorting
$json.

I'm pretty confident in the python fix, but would like extra attention
on the js fix. Go `json.Marshal()` appears to have its own sorting
rules (per the struct it's sorting? [1] [2]); I'm not sure if it's going
to match js and python.

[1] https://stackoverflow.com/questions/18668652/how-to-produce-json-with-sorted-keys-in-go
[2] https://golang.org/pkg/encoding/json/#Marshal

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,9 @@
   },
   "author": "",
   "license": "MPL-2.0",
-  "dependencies": {},
+  "dependencies": {
+    "json-stable-stringify": "^1.0.1"
+  },
   "devDependencies": {
     "assume": "^1.5.2",
     "browserify": "^14.5.0",
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "assume": "^1.5.2",
     "browserify": "^14.5.0",
     "eslint-config-taskcluster": "^3.0.0",
+    "json-stable-stringify": "^1.0.1",
     "mocha": "^4.0.1",
     "source-map-support": "^0.5.0",
     "timekeeper": "^2.0.0"
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "assume": "^1.5.2",
     "browserify": "^14.5.0",
     "eslint-config-taskcluster": "^3.0.0",
-    "json-stable-stringify": "^1.0.1",
     "mocha": "^4.0.1",
     "source-map-support": "^0.5.0",
     "timekeeper": "^2.0.0"
--- a/README.md
+++ b/README.md
@@ -183,9 +183,10 @@ calculated value to eval is not allowed.  For example, `{$eval: {$eval:
 
 ### `$json`
 
-The `$json` operator formats the given value as JSON. It does not evaluate the
-value (use `$eval` for that). While this can be useful in some cases, it is an
-unusual case to include a JSON string in a larger data structure.
+The `$json` operator formats the given value as JSON with sorted keys. It does
+not evaluate the value (use `$eval` for that). While this can be useful in some
+cases, it is an unusual case to include a JSON string in a larger data
+structure.
 
 ```yaml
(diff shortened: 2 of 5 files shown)


escapewindow:
We probably want a test to make sure an unsorted json dictionary is sorted the same across all 3 implementations.

escapewindow:
I think this is ready for review.

package.json
```diff
@@ -18,11 +18,14 @@
...
   "license": "MPL-2.0",
-  "dependencies": {},
+  "dependencies": {
+    "json-stable-stringify": "^1.0.1"
+  },
   "devDependencies": {
     "assume": "^1.5.2",
     "browserify": "^14.5.0",
     "eslint-config-taskcluster": "^3.0.0",
+    "json-stable-stringify": "^1.0.1",
```
jonasfj: having it at dependency is enough, we don't need it as `devDependency` (I'm guessing you forgot to remove this one)

jonasfj:
> Go json.Marshal() appears to have its own sorting rules (per the struct it's sorting? [1] [2])

Since all values in the golang implementations are of types: `map[string]interface{}`, `[]interface{}`, `float64`, `string`, `bool`, we have no structs... Thus, we only have to consider maps with string keys..

from [docs](https://golang.org/pkg/encoding/json/#Marshal):

>Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler. The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:
> - string keys are used directly
> - encoding.TextMarshalers are marshaled
> - integer keys are converted to strings

Again, given that we only have `map[string]interface{}` that is marshaled as objects, we only have consider strings... The _UTF-8 coercion_ mentioned is as far as I can tell:

> String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune. The angle brackets "<" and ">" are escaped to "\u003c" and "\u003e" to keep some browsers from misinterpreting JSON output as HTML. Ampersand "&" is also escaped to "\u0026" for the same reason. This escaping can be disabled using an Encoder that had SetEscapeHTML(false) called on it.

So if we do the trick with using an Encoder where we call `SetEscapeHTML(false)` then we can probably avoid most of the issues with UTF-8 in golang... In fact from the [source](https://golang.org/src/encoding/json/encode.go?#L674) it seems it just resolves to sorting lexigraphically...

To be perfectly clear the [golang specification](https://golang.org/ref/spec#Comparison_operators) defines this as:
> String values are comparable and ordered, lexically byte-wise.

----
I think we should have tests that unicode keys are also sorted correctly. But I'm 100% okay, deferring that to another bug. Just file an issue, and we can fix it later -- there is potentially a lot of particularities to work out.

jonasfj:
For the record, this is a really good catch... we should make sure json-e is functional and consistent between all implementations.

escapewindow:
Awesome, thank you!
When this merges, I'd like it rolled out everywhere that spawns decision, cron, or action tasks. I imagine treeherder will pick it up within a week; mozilla-taskcluster should hopefully be an easy update.

djmitche:
v2.5.0

djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1428889 for in-tree changes.

djmitche:
https://github.com/taskcluster/mozilla-taskcluster/pull/123 for mozilla-taskcluster

djmitche:
https://github.com/taskcluster/taskcluster-tools/pull/420 for tc-tools.

Issue #222, opened 2018-01-08
make sure $json sorts unicode strings the same across json-e implementations


escapewindow:
Per https://github.com/taskcluster/json-e/pull/221#issuecomment-355865572 :

> I think we should have tests that unicode keys are also sorted correctly. But I'm 100% okay, deferring that to another bug. Just file an issue, and we can fix it later -- there is potentially a lot of particularities to work out.

escapewindow:
Per https://github.com/taskcluster/json-e/pull/221#pullrequestreview-87227483 : 

>  I think we should take "reasonable" measures to make $json output equivalent, but where that's impossible I'm also OK with a caveat in the docs that $json may introduce non-equivalent results. That's already the case for floating-point arithmetic, I believe.

djmitche:
BTW, I said "reasonable" just to exclude heroic things like writing our own JSON encoder or some-such.  I think we can get somewhere between 99% and 100% fidelity between languages with no heroism.

djmitche:
The task here is to check that unicode characters of all sorts (astral plane, etc.), when used as property names and passed to `{$json: ..}`, produce the same output.

Biboswan:
For the following test templates we get same result in both javascript and python implementations.

template: {$json: {'\u3347\u0050': 1, '\uE1FF\u0010': 2, '\u0055\u0045': 3}}
template: {$json: {'u005A\u10000\u0050': 1, '\u{E1FFF}\u0010': 2, '\u0055a\u0045': 3}}
template: {$json: {'e\u{E3FFE}\u732b':'hey', '\u0041\u0012\u107000':'hi', 'A\u{24000}':'hello'}}
template: {$json: {'e\u{E3FFE}\u732b':'hey', '\u0041\u{24000}\u107000':'hi', 'A\u4FFF':'hello'}}
Any  more suggestions?

But they don't seem to  be sorted especially when code belongs to astral planes.
However when I used console.log to display the result it displays in expected sorted order.

djmitche:
What is the expected sort order?  I'm not familiar with Unicode sorting, honestly.  And, if we're pretty confident that the languages all produce the same output, I'm not too concerned with whether that's sorted or not.

djmitche:
I don't think the console sorts its results:
```
x
Object { x: 1, a: 2, aa: 3 }
```
so that's probably not the best thing to compare to.

[...]

PR #224, opened 2018-01-16
$json sorting unicode strings (issue #222)

Biboswan:


Biboswan:
@djmitche  add the test cases

djmitche:
```
     Unknown assertation failure occured, assumed `'{"A\\u0012ၰ00":"hi","A𤀀":"hello","e󣿾猫":"hey"}'`
to deeply equal `'{"\\u0041\\u0012\\u107000":"hi", "A\\U00024000":"hello", "e\\U000E3FFE\\u732b":"hey"}'`
```

djmitche:
I think you need to use single-quotes here, so that the `\U` are interpreted:
https://github.com/Biboswan/json-e/blob/d5ab740940818fadb759d99e3e18321d09205a07/specification.yml#L597
```
result: '{"\u0055\u0045": 3, "\u3347\u0050": 1 , "\uE1FF\u0010": 2}'
```
should be
```
result: "{\"\u0055\u0045\": 3, \"\u3347\u0050\": 1 , \"\uE1FF\u0010\": 2}"
```

Biboswan:
if 
````
 result:{\"\u0055\u0045": 3, \"\u3347\u0050": 1 , \"\uE1FF": 2}
````
then result becomes
````
`{ '\\"\\u0055\\u0045"': 3, '\\"\\u3347\\u0050"': 1, '\\"\\uE1FF"': 2 }`
````

If just quotes around the object is removed i.e
````
{"\u0055\u0045": 3, "\u3347\u0050": 1 , "\uE1FF": 2}
````
then
````
result :{ UE: 3, '㍇P': 1, '': 2 }
````

But necessary double quotes are also missing around the keys. Using more quotes around keys prevents unicode to be interpreted.
Seems like only some parser can handle this.

**Comment by djmitche** [comment]
I'll see if I can get it to work.  If not, maybe we can add some special-casing to the tests.

**Comment by djmitche** [comment]
This can get a little tricky since we need to write the expected values as an escaped string for the YAML parser.  That means we need to use double-quotes (`result: ".."`) and then escape `"` and `\` in the string itself when we want them to appear literally.  We can use `\uXXXX` and `\Uxxxxxxxx` for unicode characters.  Also, whitespace matters in the JSON output. With all of that fixed, the following works in JS:

```yaml
---
title:    sorting pairs by unicode key strings sorts lexically by codepoint
context:  {}
template: {$json: {"\u3347\u0050": 1, "\uE1FF\u0100": 2, "\u0055\u0045": 3}}
result:   "{\"\u0055\u0045\":3,\"\u3347\u0050\":1,\"\uE1FF\u0100\":2}"
---
context:  {}
result:   "{\"UaE\":3,\"Z\U00010000P\":1,\"\U000E1FFF\u0103\":2,\"\U0001F809\u732b\":\"hey\"}"
```

But it fails in Python:

```
======================================================================
FAIL: test.test_specification.test('$json: sorting pairs by unicode key strings sorts lexically by codepoint',)
----------------------------------------------------------------------
  File "/home/dustin/go/src/github.com/taskcluster/json-e/.eggs/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/dustin/go/src/github.com/taskcluster/json-e/test/test_specification.py", line 29, in test
    '{!r} != {!r}'.format(res, spec['result'])
AssertionError: u'{"UE":3,"\\u3347P":1,"\\ue1ff\\u0100":2}' != u'{"UE":3,"\u3347P":1,"\ue1ff\u0100":2}'

======================================================================
FAIL: test.test_specification.test('$json: sorting pairs by unicode key strings sorts lexically by codepoint, even with chars above base plane',)
----------------------------------------------------------------------
  File "/home/dustin/go/src/github.com/taskcluster/json-e/.eggs/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/dustin/go/src/github.com/taskcluster/json-e/test/test_specification.py", line 29, in test
    '{!r} != {!r}'.format(res, spec['result'])
AssertionError: u'{"UaE":3,"Z\\ud800\\udc00P":1,"\\udb47\\udfff\\u0103":2,"\\udb4f\\udffe\\u732b":"hey"}' != u'{"UaE":3,"Z\U00010000P":1,"\U000e1fff\u0103":2,"\U0001f809\u732b":"hey"}'
```

So let's back up a little bit and write some tests for representations of unicode characters in $json output, before we start worrying about sort order.

```yaml
context:  {}
```

This passes in golang and in JS, but not in Python.  It expects a `\u`-escaped string, an ASCII "Z", a literal LATIN SMALL LETTER A WITH BREVE (codepoint 0103), and a literal UPWARDS ARROW WITH LARGE TRIANGLE ARROWHEAD (codepoint 0001f809).

Python gives:

```
AssertionError: u'["\\u0010","Z","\\u0103","\\ud83e\\udc09"]' != u'["\\u0010","Z","\u0103","\U0001f809"]'
```

here Python has the same result for the first two characters, but uses a `\u`-escaped 0103 and, worse, represents codepoint 0001f809 as the `\u`-escaped surrogate pair d83e-dc09.  It looks like Python is only using ASCII characters in the output, with `\u` escapes for anything non-ASCII.  @indygreg helpfully pointed me at the `ensure_ascii` to `json.dump`, which defaults to True.  Setting that to False seems to fix everything!

djmitche:
I pushed my changes to this PR.  Let me know what you think?

djmitche:
It looks like the golang failures are typical metalinter failures and not related to this patch:

```
jsone.go:1::warning: file is not gofmted with -s (gofmt)
jsone.go:1::warning: file is not goimported (goimports)
```

djmitche:
Whoops, I meant to squash that!

PR #227, opened 2018-01-19
Gometalinter config

jonasfj:
based on #226


.taskcluster.yml
```diff
@@ -150,7 +150,10 @@ tasks:
           git config advice.detachedHead false &&
           git checkout {{event.head.sha}} &&
           go get -v -d -t ./... &&
-          go test -v -race ./...
+          go test -v -race ./... &&
+          go get -u github.com/alecthomas/gometalinter &&
```
djmitche: Can we pin a version here somehow?  It seems like this starts failing when we get a newer version of the tool (or one of its dependencies?)

jonasfj: I don't have a good way of doing this... I'll file an issue for it..

jonasfj: Filed #228, I agree we should find a good way to do this.

PR #229, opened 2018-02-01
Use Towncrier to create Changelog

### LINKED ISSUE #193: Missing release notes
I couldn't find a CHANGES.md, or section in the README, or release notes added via the GitHub 'releases' feature.

I don't suppose some could be added for future releases?

djmitche:
Fixes #193.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,51 @@
+# JSON-e development
+
+JSON-e is implemented in several languages. All implementations should behave
+identically, and the test suite subjects each implementation to the same
+requirements (in `specification.yml`). Developing JSON-e will usually require
+you to make changes in all implementations.
+
+You can run `./test.sh` to run json-e's tests and lint checks across all
+implementations.
+
+## JavaScript
+
+You should run `npm install` to install the required packages for JSON-e's
+execution and development.
+
+## Python
+
+For Python, activate a virtualenv and run `pip install -e .`.
+
+To run the Python tests only, use `python setup.py test`.
+
+Note that JSON-e supports both Python 2 and Python 3.
+
+# Demo development
+
+The demo website is a [Neutrino](https://neutrino.js.org/) app hosted in
+`demo/`.  Follow the usual Neutrino development process (`yarn install && yarn
+start`) there.
+
+The resulting application embeds and enriches the README file.
+
+# Changelog
+
+When making a pull request with your changes, create a new file in
+`newsfragments/` named after the issue or bug you are working on, with a suffix
+of `.bugfix`, `.feature`, or (for docs-only changes) `.doc`.  The content of
+this file should be in reStructuredText format. Keep it to a simple sentence,
+and you should be fine!
+
+For example, `newsfragments/201.bugfix` might contain `Fixed the precedence of
+the || and 'in' operators`.
+
+# Making a Release
+
+* Run `towncrier --version=$newversion --draft` and check that the output looks OK.  Then run it without `--draft`, deleting the old newsfiles.  Commit.
+* Update the version, commit, and tag -- `npm version patch` (or minor or major, depending)
+* Push to release the JS version -- `git push && git push --tags`
+* Release to PyPi:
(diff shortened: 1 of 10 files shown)


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,51 @@
...
+
+# Making a Release
+
+* Run `towncrier --version=$newversion --draft` and check that the output looks OK.  Then run it without `--draft`, deleting the old newsfiles.  Commit.
+* Update the version, commit, and tag -- `npm version patch` (or minor or major, depending)
+* Push to release the JS version -- `git push && git push --tags`
+* Release to PyPi:
+  * `python setup.py sdist`
+  * `twine upload dist/json-e-<version>.tar.gz`
+* Go doesn't believe in versions, so there's notihng to do for a Go release!
```
jonasfj: For golang we could use http://gopkg.in` and then tags in github...

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,51 @@
...
+of `.bugfix`, `.feature`, or (for docs-only changes) `.doc`.  The content of
+this file should be in reStructuredText format. Keep it to a simple sentence,
+and you should be fine!
+
+For example, `newsfragments/201.bugfix` might contain `Fixed the precedence of
+the || and 'in' operators`.
+
+# Making a Release
+
+* Run `towncrier --version=$newversion --draft` and check that the output looks OK.  Then run it without `--draft`, deleting the old newsfiles.  Commit.
```
jonasfj: This is too complicated, make it a script and list the credentials that you must have in-order to make a release...

Or even better, let's make travis responsible for releases and just make releasing the result of pushing a tag...
Or running a script that does `towncrier` and pushes a tag...

jsone/newsfragments/193.feature
```diff
@@ -0,0 +1 @@
+Begin using towncrier to build CHANGELOG.rst
```
jonasfj: I'm not a fan... I would much prefer some test logic that enforced sane commit messages... And then we build changelog from commit messages when we make a new release...

Also why is it in the `jsone` sub-folder? I'm not sure where it should be? and should it be called `newsfragments/` why not `.changelog/`.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,51 @@
...
+start`) there.
+
+The resulting application embeds and enriches the README file.
+
+# Changelog
+
+When making a pull request with your changes, create a new file in
+`newsfragments/` named after the issue or bug you are working on, with a suffix
+of `.bugfix`, `.feature`, or (for docs-only changes) `.doc`.  The content of
+this file should be in reStructuredText format. Keep it to a simple sentence,
```
jonasfj: Why not markdown?

Couldn't we just have a script pull it from github commit log... like: https://github.com/conventional-changelog/standard-version

djmitche:
@edmorley made some good arguments against considering the git log to be a changelog, and the towncrier README does, too:

> towncrier delivers the news which is convenient to those that hear it, not those that write it.
> 
> That is, by duplicating what has changed from the “developer log” (which may contain complex information about the original issue, how it was fixed, who authored the fix, and who reviewed the fix) into a “news fragment” (a small file containing just enough information to be useful to end users), towncrier can produce a digest of the changes which is valuable to those who may wish to use the software. These fragments are also commonly called “topfiles” or “newsfiles” in Twisted parlance.

The limitations here (reStructuredText, calling it newsfragments, putting it in the python package directory) are just the way towncrier works.  My intent here was to do the simplest thing that would make @edmorley happy, not to embark on a project :)

@jonasfj it sounds like you're volunteering to set this up with another tool, with some on-tag automation, maybe some fanciness to create new release objects in github.  Can I assign #193 to you?

jonasfj:
> My intent here was to do the simplest thing that would make @edmorley happy, not to embark on a project :)

Fair point, I might be over thinking this...

jonasfj:
So IMO, we should still script the release process... but that can be a separate issue...

Issue #238, opened 2018-04-12
CLI tool for JSON-e


wryun:
Just wrote:

https://github.com/wryun/rjsone

Is it worth PR-ing (with some tests, perhaps...)? Or maybe you could link to it on the website?

djmitche:
That's really cool!

I think it makes sense as a separate project more than one incorporated in this repo.  But I'd love to have a link to it in the docs in this repo.  Do you want to make a PR to add that?

wryun:
Thanks!

PR #239, opened 2018-05-22
Link to rjsone in README

wryun:
Link to command line rjsone tool

See #238


wryun:
Thanks for the super-fast response and merge!

Issue #235, opened 2018-04-03
Get index in map


shaikshakeel:
in map function 

we can apply our custom transformation to each value of map(either array element or JSON key, value)

is there any possibility  to get index of element in case arrays ?
```
template:
  $map: [2, 4, 6]
  each(x,index): {$eval: 'x + index'}
context:  {a: 1}
result:   [2, 5, 8]

```

djmitche:
This sound OK to me.  Do you want to implement it?

shaikshakeel:
Mostly, trying to understand how map works in json-e :)

PR #242, opened 2018-06-20
Add map index support (js/python)

cronik:
Extend the $map operation to expose the value index as specified in #235 . An optional argument can be supplied to the "each" argument (i.e. each(val,index)). If no index argument is specified an implicit '_index' value will be added to the context.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -10,16 +10,31 @@ implementations.
 
 ## JavaScript
 
+The JavaScript implementation is located in `./src`
+
 You should run `npm install` to install the required packages for JSON-e's
 execution and development.
 
+To run the tests use `npm test`
+
 ## Python
+
+The Python implementaion is located in `./jsone`.
 
 For Python, activate a virtualenv and run `pip install -e .`.
 
 To run the Python tests only, use `python setup.py test`.
 
 Note that JSON-e supports both Python 2 and Python 3.
+
+## Go
+
+The Go implementation is located in `./jsone.go` and `./interpreter`.
+
+To run the Go test run
+```
+GOPATH=$(pwd) go test -v -race ./...
+```
 
 # Demo development
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -31,6 +31,11 @@ Note that JSON-e supports both Python 2 and Python 3.
 
 The Go implementation is located in `./jsone.go` and `./interpreter`.
 
+Install dependencies:
+```
+GOPATH=$(pwd) go get -t ./...
+```
+
 To run the Go test run
 ```
 GOPATH=$(pwd) go test -v -race ./...
--- a/interpreter/util.go
+++ b/interpreter/util.go
@@ -242,7 +242,7 @@ func IsValidData(data interface{}) error {
 		return nil
 	}
 	switch val := data.(type) {
-	case *function, string, int, float64, bool:
+	case *function, string, float64, bool:
 		return nil
 	case []interface{}:
 		for _, value := range val {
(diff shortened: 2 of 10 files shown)


cronik:
@djmitche how about each(value, key) for objects. This will make it consistent with other libraries like [lodash](https://lodash.com/docs/4.17.10#forEach) and [underscore](https://underscorejs.org/#each). 

To maintain backward compatibility, if they don't specify 2 arguments we pass `{key, val}`.
```yml
---
title:    $map each 2 args
context:  {}
template:
  $map: {a: 1, b: 2, c: 3}
  each(k,v): {'${k}x': {$eval: 'v + 1'}}
result: {ax: 2, bx: 3, cx: 4}
---
title:    $map each single arg
context:  {}
template:
  $map: {a: 1, b: 2, c: 3}
  each(y): {'${y.key}x': {$eval: 'y.val + 1'}}
result: {ax: 2, bx: 3, cx: 4}
```

cronik:
@djmitche I'm not very familiar with go, but I'll give it a shot. Are any project setup instructions for 'go'? I get the follow error when trying to run the tests:

```
➜  json-e git:(map-index) ✗  go test -v -race ./...
jsone.go:15:2: cannot find package "github.com/taskcluster/json-e/interpreter" in any of:
        /usr/local/Cellar/go/1.10.3/libexec/src/github.com/taskcluster/json-e/interpreter (from $GOROOT)
        /Users/croninka/go/src/github.com/taskcluster/json-e/interpreter (from $GOPATH)
interpreter/interpreter.go:9:2: cannot find package "github.com/taskcluster/json-e/interpreter/prattparser" in any of:
        /usr/local/Cellar/go/1.10.3/libexec/src/github.com/taskcluster/json-e/interpreter/prattparser (from $GOROOT)
        /Users/croninka/go/src/github.com/taskcluster/json-e/interpreter/prattparser (from $GOPATH)
```

djmitche:
Hm, there's no good advice in https://github.com/taskcluster/json-e/blob/master/.github/CONTRIBUTING.md

I believe you need to run `go get -t ./...` first.  Note that that will update other things in your GOROOT (~/go) which might cause other go projects you're working on to have problems.

.github/CONTRIBUTING.md
```diff
@@ -10,17 +10,32 @@ implementations.
...
 
 To run the Python tests only, use `python setup.py test`.
 
 Note that JSON-e supports both Python 2 and Python 3.
 
+## Go
+
+The Go implementation is located in `./jsone.go` and `./interpreter`.
+
+To run the Go test run
```
djmitche: did you want to mention the `go get` here, too?

interpreter/util.go
```diff
@@ -242,7 +242,7 @@ func IsValidData(data interface{}) error {
 		return nil
 	}
 	switch val := data.(type) {
-	case *function, string, float64, bool:
+	case *function, string, int, float64, bool:
```
djmitche: Hm, this is an interesting change.  Is this here because the index value is an int, not a float64?  I suspect we should be using float64 everywhere for what JSON defines as a "Number".  @imbstack what do you think?

imbstack: Yeah, I think we need to enforce that everything is `float64`. I agree that makes indexes seem weird...

Maybe so long as we render everything as a `float64` it is ok? We can be accepting of number-like inputs but only write JSON compatible numbers?

djmitche:
@cronik did you want to wrap this up?  See @imbstack's comments..

cronik:
@djmitche and @imbstack, I'm not clear what changes are requested. Should the index be cast as a `float64`?

djmitche:
Yes, I think that's the only remaining change desired.

djmitche:
Thanks @cronik!

PR #245, opened 2018-06-28
Document `typeof()`

djmitche:
I'm not totally happy about the null handling, but it seems like we should err on the side of not changing things..


djmitche:
I'm really impressed that even without the test, all three languages pass..

PR #248, opened 2018-07-06
Include guidance on JS installation and use in browsers

### LINKED ISSUE #247: How do I use it in NodeJs or Browser??
If I want to use in NodeJs
what's the NPM package name ?
What method/class is exposed by the package?

If I want to use in browser what file to include?

Nothing is documented any where except mentioning "import jsone from 'json-e';"
An example for working in browser and nodejs would be nice,

djmitche:
Fixes #247.


djmitche:
Updated with that info -- thanks!

eliperelman:
LGTM

PR #251, opened 2018-08-07
Add `$match` operator

owlishDeveloper:
See discussion here https://github.com/taskcluster/json-e/issues/161

# Checklist

* [x] Added a short description of the change to `newsfiles/$issue.bugfix` (for fixes) or `.feature` (for new features) or `.doc` (for documentation-only changes)

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,11 +260,7 @@ template: {$match: [{"x == 10": "ten"}, {"x == 20": "twenty"}]
 context: {x: 10}
 result: ["ten"]
 ```
-```yaml
-template: {$match: {"x == 10": "ten", "x == 20": "twenty"}
-context: {x: 10}
-result: {"x": "ten"}
-```
+
 ```yaml
 template: {$match: [{"x == 10 || x == 20": "tens"}, {"x == 10": "ten"}]
 context: {x: 10}
--- a/README.md
+++ b/README.md
@@ -251,6 +251,26 @@ context: {cond: false}
 result: {other: 3}
 ```
 
+### `$match`
+
+The `$match` operator supports pattern matching.
+
+```yaml
+template: {$match: [{"x == 10": "ten"}, {"x == 20": "twenty"}]
+context: {x: 10}
+result: ["ten"]
+```
+```yaml
+template: {$match: {"x == 10": "ten", "x == 20": "twenty"}
+context: {x: 10}
+result: {"x": "ten"}
+```
+```yaml
+template: {$match: [{"x == 10 || x == 20": "tens"}, {"x == 10": "ten"}]
+context: {x: 10}
+result: ["tens", "ten"]
+```
+
 ### `$flatten`
 
 The `$flatten` operator flattens an array of arrays into one array.
--- a/README.md
+++ b/README.md
@@ -348,23 +348,24 @@ result: {ax: 2, bx: 3, cx: 4}
 
 ### `$match`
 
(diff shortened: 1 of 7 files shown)


owlishDeveloper:
I have some questions:

I can't find `newsfiles` anywhere 🤔😳 Should I correct the template?

I was thinking what to do with the situation `{'cond == 3': 4, 'cond < 5 && cond > 0': 4}` (several conditions in one object). Like, in ideal world, I would imagine `$match` to be an array of objects with exactly one condition each. Disallowing multiple conditions seemed restrictive, but at the same time it opens the gates for mess, namely, the order of the results will be arbitrary (because some languages randomize map iteration like Go, some sort the entries like Python 2.7, some just stick to the order in which the entries were added or declared, like JavaScript or Python 3.5).

Between restriction and mess, I chose mess. What do you think?

owlishDeveloper:
@djmitche @imbstack please review 😊

imbstack:
I think the directory is `newsfragments` now! If you can update the template that would be great.

specification.yml
```diff
@@ -277,6 +277,29 @@ template: {$if: 'cond', then: 1, foo: 'bar', else: 2, bing: 'baz'}
...
+context:  {cond: 3}
+template: {$match: [{'cond > 5': 2}, {'cond == 5': 3}, {'cond == 4 || cond < 1': 4}]}
+result:   []
+---
+title:    $match, is not an array
+context:  {cond: 3}
+template: {$match: {'cond < 5 && cond > 0': 2}}
+error:    'TemplateError: $match can evaluate arrays only'
+---
+title:    $match, for many true conditions in 1 object order of results is arbitrary. The whole array is flat
```
imbstack: If this is desired behavior, can we add it to the readme?

Also let's add a test here for when the value is something that needs to be evaluated itself.

README.md
```diff
@@ -346,6 +346,27 @@ context:  {}
...
+```yaml
+template: {$match: [{"x == 10 || x == 20": "tens"}, {"x == 10": "ten"}]
+context: {x: 10}
+result: ["tens", "ten"]
+```
+```yaml
+template: {$match: [{"x == 10 || x == 20": "tens", "x == 10": "ten"}]
+context: {x: 10}
+result: ["tens", "ten"]
+```
```
djmitche: What is the rationale for supporting both of these, if they do the same thing?  It seems like the simplest way to express this would be

```yaml
template: {$match: {"x == 10 || x == 20": "tens", "x == 10": "ten"}}
context: {x: 10}
result: ["tens", "ten"]
```

what is the benefit of the `[..]` in the template?

specification.yml
```diff
@@ -831,6 +836,69 @@ template:
...
+---
+title:    $match, 2 matches
+context:  {cond: 3}
+template: {$match: [{'cond == 3': 2}, {'cond == 5': 3}, {'cond < 5 && cond > 0': 4}]}
+result:   [2, 4]
+---
+title:    $match, no matches
+context:  {cond: 3}
+template: {$match: [{'cond > 5': 2}, {'cond == 5': 3}, {'cond == 4 || cond < 1': 4}]}
+result:   null
```
djmitche: This should be in the README, too -- in the absence of other docs I would expect no matches to return `[]`.

owlishDeveloper: I am actually thinking that I'll probably return to an empty array. That way we can always map over results etc., and no need for `null` checks. What do you think?

README.md
owlishDeveloper: Yes, definitely looks better with no array! Thank you 😊

PR #259, opened 2018-09-21
Ensure that order of $match's result is deterministic.

djmitche:
It doesn't really matter what the order is, but it should be the same for all implementations.

# Checklist


owlishDeveloper:
Why are the go tests failing? Weird, lemme see

djmitche:
Looking.  Go tests passed for me locally..

djmitche:
Oh, it's lint:
```
jsone.go:1::warning: file is not gofmted with -s (gofmt)
jsone.go:555:18:warning: should omit 2nd value from range; this loop is equivalent to `for condition := range ...` (golint)
jsone.go:572:5:warning: should write value := match[key] instead of value, _ := match[key] (S1005) (gosimple)
```
which is weird b/c vim should be gofmt'ing for me.  But I'll fix.

djmitche:
2.7.1

PR #261, opened 2018-10-30
Make typeof(null) be null, not null

### LINKED ISSUE #246: Make typeof(null) be "null", not null
See https://github.com/taskcluster/json-e/pull/245 where it turns out that this function has some unfortunate behaviors.

alexandrasp:
I change typeof functionality in builtins.js to return null as a string. Fix #246 

# Checklist

### DIFF
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -43,7 +43,7 @@ def interpolate(string, context):
             if isinstance(parsed, (list, dict)):
                 raise TemplateError(
                     "interpolation of '{}' produced an array or object".format(string[:offset]))
-            if to_str(parsed) == "null":
+            if parsed is None:
                 result.append("")
             else:
                 result.append(to_str(parsed))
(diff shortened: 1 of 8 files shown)


djmitche:
Good start -- this will need changes in `specification.yml`, and there are three more languages to fix :)

jsone/render.py
```diff
@@ -43,7 +43,7 @@ def interpolate(string, context):
             if isinstance(parsed, (list, dict)):
                 raise TemplateError(
                     "interpolation of '{}' produced an array or object".format(string[:offset]))
-            if to_str(parsed) == "null":
+            if to_str(parsed) == "null" and parsed != "null":
```
alexandrasp: really hard figure out that parse should be different of `null` here!

djmitche: There are two cases where `to_str(parsed)` is "null": `parsed=None` and `parsed="null"`.  So I think this statement is equivalent to `if parsed is None`.  To check, I made that change and ran the tests and they still passed.

djmitche:
Released in v3.0.0

Issue #266, opened 2019-01-28
Remove usage of Node built-in 'assert'


edmorley:
Hi!

Treeherder uses `json-e` in the browser, for the taskcluster actions feature - eg:
https://github.com/mozilla/treeherder/blob/23036375f9faec4c0f343e51024eb7feb2195c45/ui/models/taskcluster.js#L41

In the upcoming webpack 5, webpack no longer automatically polyfills node built-ins in imported libraries in an effort to reduce unintended bundle size bloat.

Trying webpack 5 out against Treeherder I get several errors like:

```
ERROR in ./node_modules/json-e/src/index.js 3:13-30
Module not found: Error: Can't resolve 'assert' in 'C:\Users\Ed\src\treeherder\node_modules\json-e\src'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need these module and configure a polyfill for it.
```

Whilst we could configure a polyfill, I think it would be best for json-e to stop using `assert` so it's browser-compatible out of the box.

Currently it's used in the following locations (excluding tests):
https://github.com/taskcluster/json-e/blob/24a97f3474bc4e9625b17d3b6e91f9ff99eef380/src/index.js#L3
https://github.com/taskcluster/json-e/blob/5f1fefda8b07ed0016c98f77cd4640ec8a920201/src/tokenizer.js#L40-L47
https://github.com/taskcluster/json-e/blob/5f1fefda8b07ed0016c98f77cd4640ec8a920201/src/prattparser.js#L39

Many thanks :-)

edmorley:
Thank you :-)

PR #268, opened 2019-01-28
remove uses of `assert` in non-testing code

djmitche:
Per #266, we should avoid using `assert` in production code, as it is not defined on browsers.

### DIFF
--- a/jsone/newsfragments/266.bugfix
+++ b/jsone/newsfragments/266.bugfix
@@ -1,0 +1,1 @@
+Do not import `assert` in non-test code, so that it can run on browsers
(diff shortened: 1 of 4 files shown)


djmitche:
CI failure is in Go.

djmitche:
#269 for the Go failures

edmorley:
> I wonder if we can ban assert with a lint?

I'm pretty sure eslint would pick this up so long as `envs` only includes `browser` and not `node`.

Currently this repo uses an eslint config that sets `node`:
https://github.com/taskcluster/eslint-config-taskcluster/blob/dc1bd87514b5a2bd0fe3fa017e70f82823f235cf/index.js#L3

imbstack:
Thanks @edmorley. I've gone ahead and updated our lint in taskcluster/json-e#270

Issue #269, opened 2019-01-28
Switch from gometalinter to golangci-lint


djmitche:
gometalinter changes all the time, meaning that it starts failing for silly reasons (like adding or removing a linter) and causes PRs to go red.  golangci-lint seems a lot more stable (it's what I'm using for websocktunnel).  Let's switch to that.

PR #270, opened 2019-01-29
Update to more reasonable linting

imbstack:
This should address issues we had with using assert in the library among
other things.


src/interpreter.js
```diff
@@ -148,17 +148,11 @@ let testMathOperands = (operator, left, right) => {
...
-  if (!isBool(operand)) {
-    throw expectationError(`infix: ${operator}`, `boolean ${operator} boolean`);
-  }
-};
-
 let prefixRules = {};
 let infixRules = {};
 
 // defining prefix rules
-prefixRules['number'] = (token, ctx) => {
```
djmitche: I dislike this mildly -- it makes it much harder for someone to know, later, what arguments this function actually gets.

.gometalinter.json
```diff
@@ -4,7 +4,6 @@
     "gofmt",
     "goimports",
     "golint",
-    "gosimple",
```
djmitche: #269

src/interpreter.js
imbstack: Yeah, agreed. I can turn on the `after-used` option  for this and add them back.

PR #286, opened 2019-04-19
Update README.md to document the Pypi distribution name

Lothiraldan:
As the module name `jsone` doesn't yields the distribution on Pypi when
searched, document the actual distribution name to ease installation.


djmitche:
Thank you!

PR #288, opened 2019-04-24
Adds python_requires to setup.py

### LINKED ISSUE #287: [python] Define python_requires in setup.py to show compatibility with both py2 and py3
I've discovered that the `json-e` library doesn't support Python 3 - for example, when running `jsone.render(...)`, the strings in the task payloads had byte prefixes in the strings themselves, such as: `"b'A1k6Dd-jRr-KQnccuFjc2A'"`.

Perhaps we can use the [`python_requires`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) property in `setup.py` to document that the `json-e` package is specific to Python 2?

mitchhentges:
Fixes #287

# Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,5 +36,6 @@ setup(name='json-e',
         "PyYAML",
         "python-dateutil",
         'pep8',
-    ]
+    ],
+    python_requires='~=2.7, >=3.6'
 )
--- a/jsone/newsfragments/287.bugfix
+++ b/jsone/newsfragments/287.bugfix
@@ -1,0 +1,1 @@
+Add `python_requires` to package definition


setup.py
```diff
@@ -36,5 +36,6 @@
         "PyYAML",
         "python-dateutil",
         'pep8',
-    ]
+    ],
+    python_requires='>=2'
```
mitchhentges: I'm not sure if it's preferred that the exact `.x` be added here, like `2.7` :thinking:

djmitche: I think it goes without saying that we don't support Python-1.x (the last version of which was released in the previous century).  I'd suggest looking at the `.taskcluster.yml` and indicating the Python versions we test on.  We support 2.7 and some minimum version of 3.x, but not 3.y for y < x.

mitchhentges: Your python 3 tests in `.taskcluster.yml` only test python `3.6`: I can enforce that as a minimum, but it might be overly restrictive (`json-e` might still work for `3.5`, but just might be untested, just like how `3.7` is untested).

I'll update the PR accordingly, but it might cause breakage downstream

djmitche: We can certainly add testing for those versions, to check.

Note, too, that this PR failed its Python tests :)

mitchhentges:
I think that the `python_requires` string that most accurately _matches_ the existing (lack of) definition is `>= 2.7`, but "no documentation is better than bad documentation", so if there's an issue with, say, `3.4`, then that string would be incorrect.

> We can certainly add testing for those versions, to check.

The scope of this PR is increasing significantly.
As a non-primary `json-e` developer, I don't want to make assumptions about which versions of Python you'd want to support (just down to `3.5`? Earlier than that?), or how you want to handle downstream breakage caused by formalizing your "supported python versions".

Since the bulk of the work here is larger-context decision-making for this specific project, I'll step down this PR

Issue #294, opened 2019-06-04
I'm using this package with TS and noticed there's no type definition (index.t.ds) file for the npm package


siwalikm:
There's no type definition for this package, would you be interested in adding one if I raised a PR for the same? @djmitche

djmitche:
Sounds fine to me!

PR #303, opened 2019-06-30
Adding TS type definitions for json-e project

siwalikm:
*Adding TS type definitions for json-e project wrt. to issue #294*

- updated Readme with information on usage
- updated package.json file with url of types file
- added module definition for json-e

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ npm install --save json-e
 yarn add json-e
 ```
 
-Note: Type definations are included with this package, so there's no need of seperate
+Note: Type definitions are included with this package, so there's no need of seperate
 `@types/..` installation.
 
 As 'json-e' is a CommonJS module, the package must be imported like this [(more..)](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) for type definitions to work properly:
--- a/README.md
+++ b/README.md
@@ -71,6 +71,29 @@ tool that understands CommonJS such as Webpack.  See
 [Neutrino](https://neutrino.js.org/) for an easy, configuration-free way to
 build such applications.
 
+## TypeScript
+
+The JS module is installed with either of
+
+```shell
+npm install --save json-e
+yarn add json-e
+```
+
+Note: Type definations are included with this package, so there's no need of seperate
+`@types/..` installation.
+
+As 'json-e' is a CommonJS module, the package must be imported like this [(more..)](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) for type definitions to work properly:
+
+```typescript
+import jsone = require('json-e');
+
+var template = {a: {$eval: "foo.bar"}};
+var context = {foo: {bar: "zoo"}};
+console.log(jsone(template, context));
+// -> { a: 'zoo' }
+```
+
 ## Python
 
 The Python distribution is installed with
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,6 +1,5 @@
 // Type definitions for json-e
 // Project: https://github.com/taskcluster/json-e
-// Definitions by: siwalik <https://github.com/siwalikm>
 
 declare module 'json-e' {
   function jsone(template: Record<any, any>, context: Record<any, any>): any;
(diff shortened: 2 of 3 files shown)


owlishDeveloper:
@siwalikm welcome to json-e project! There will be a delay in me reviewing this, and I apologize for that! I'll review on Monday.

siwalikm:
Thank you @owlishDeveloper.

README.md
```diff
@@ -71,6 +71,29 @@ tool that understands CommonJS such as Webpack.  See
...
+## TypeScript
+
+The JS module is installed with either of
+
+```shell
+npm install --save json-e
+yarn add json-e
+```
+
+Note: Type definations are included with this package, so there's no need of seperate
```
owlishDeveloper: definations ---> definitions

src/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// Type definitions for json-e
+// Project: https://github.com/taskcluster/json-e
+// Definitions by: siwalik <https://github.com/siwalikm>
```
owlishDeveloper: Please remove this line. There are many type definitions that can be added to this project, and it's possible that other people will come and contribute.

src/index.d.ts
```diff
@@ -0,0 +1,8 @@
...
+
+declare module 'json-e' {
+  function jsone(template: Record<any, any>, context: Record<any, any>): any;
```
owlishDeveloper: Out of curiosity: why did you decide to use `Record<any, any>` here? As opposed to just `any`, for example

README.md
siwalikm: Corrected the typo.

src/index.d.ts
siwalikm: Defining types of `template` and `context` as `any` would let us pass any data types to the function. eg. `jsone(12, 'hello_there')` this would qualify as valid syntax.
But as we've defined `Record<any, any>`, it would now expect **only** objects but with any key and value pairs - which is more aligned towards how the function operates.

src/index.d.ts
siwalikm: I was following the official TS def. template which lets us define author(s) [(link here)](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html). 

I've removed it now as requested.

siwalikm:
@owlishDeveloper thanks a lot for reviewing this so fast.
I've committed the changes you requested.

src/index.d.ts
owlishDeveloper: makes sense!

owlishDeveloper:
@siwalikm now that the PR is merged, I am curious if a release would be useful for you?

siwalikm:
@owlishDeveloper yes, I would need a new release.

Btw I have a concern, just noticed that latest release of the package is `v3.0.0` as of now but in npmjs.com it still shows `v2.7.1`.

I guess we have to bump up the semver minor version first in our package.json and then run `npm publish` (as the authorised npm package's author) to to update the default version of the package to get this and all latest changes.

I can raise a PR and update the minor version in package.json if you want, but would you be able to publish the updated version to npm?

djmitche:
> Btw I have a concern, just noticed that latest release of the package is v3.0.0 as of now but in npmjs.com it still shows v2.7.1.

Interesting -- 3.0.0 is on the [list of versions](https://www.npmjs.com/package/json-e) but 2.7.1 was published a few seconds later so apparently npm considers it the latest.  At any rate, publishing 3.0.1 now shouldn't be a problem.

djmitche:
(sorry to butt in, but it looks like the 3.0.0 issue was my fault..)

siwalikm:
Ah, that seems straight forward then. Can we do a 3.0.1 release please @djmitche ? 🙏

owlishDeveloper:
@siwalikm I did v3.0.1 release (it's on npm as well). Let me know if the types work for you (it seems to me there's an additional step to publish the types)

siwalikm:
@owlishDeveloper @djmitche updated my project and everything works as expected. 🙌Thanks.

gif

Issue #249, opened 2018-07-17
Allow value of $let to be evaluated


djmitche:
```yaml
$let:
  $if: something
  then: {a: 10, b: 20}
  else: {a: 20, b: 10}
in:
  $eval: 'a + b'
```

As of right now, this is not legal, because the value of the `$let` is not evaluated -- it must be an actual object, the *values* of which are evaluated.  But we could just evaluate the whole thing and require that the result be an object (which is not then re-evaluated)

terrameijar:
@owlishDeveloper I would like to try my hand at this issue. Could you please give me pointers on how to approach it? What is the best way to make sure `$let` gets evaluated?

stuartelimu:
can i work on this issue

owlishDeveloper:
@terrameijar before starting working on the issues, you have to setup your development environment, successfully run the tests and send me the screenshots of the successful test runs. Thank you!

@stuartelimu issues can be worked on only one at a time. You already have one assigned. Thank you!

owlishDeveloper:
@richardgreg did you mean you would like to work on this issue?

vishakhanihore:
@owlishDeveloper I have successfully set up the project. Can you please tell me your email-id ?

osumgbachiamaka:
@owlishDeveloper can i have this task please

vishakhanihore:
@owlishDeveloper Sent you ss, Please assign

richardgreg:
@owlishDeveloper yes, please!

owlishDeveloper:
@richardgreg asked first (terrameijar already has another issue assigned), so this issue goes to richardgreg!

richardgreg:
@djmitche pls could you provide the result for the JSON-e statement you wrote, assuming the issue has been resolved?

owlishDeveloper:
@richardgreg  in that particular example the result should be `30`. Let me amend the example a bit:

```yml
$let:
  $if: something
  then: {a: 10, b: 20}
  else: {a: 10, b: 10}
in:
  $eval: 'a + b'
```
Here, is `something` evaluates to `true`, the result should be `30`, otherwise `20`

gregsdennis:
What's the difference between this and #212?  It seems to me that implementing this issue overturns that decision.

djmitche:
That's a good point - I don't think we had considered that this decision had been made in the opposite direction already.

[...]

PR #328, opened 2019-10-07
operators: Allow $let to be evaluated.

richardgreg:
*describe your pull request here*
issue #249 
# Checklist

### DIFF
--- a/jsone/newsfragments/249.bugfix
+++ b/jsone/newsfragments/249.bugfix
@@ -1,0 +1,1 @@
+Fixed the issue of '$let' not evaluating if the key is an operator.
--- a/jsone/newsfragments/249.bugfix
+++ b/jsone/newsfragments/249.bugfix
@@ -1,1 +1,39 @@
-Fixed the issue of '$let' not evaluating if the key is an operator.
+Fixed the issue of speacial values in '$let' not evaluating.
+
+Values like 'if-then-else' statements, '$eval', and rendered keys.
+The values evaluate to an object, the values of which are evaluated.
+For example, the operations below no longer return an error:
+
+title: let with a value of $if-then-else
+$let:
+  $if: something == 3
+  then: {a: 10, b: 10}
+  else: {a: 20, b: 10}
+in:
+  $eval: 'a + b'
+
+context: {'something': 3}
+result:  20
+
+-----
+
+title: let using values from the parent context
+$let:
+  "b": {$eval: "a + 10"}
+in:
+  $eval: 'a + b'
+
+context: {'a':5}
+result:  20
+
+-----
+
+title: let with a rendered key
+$let:
+  "first_${name}": 1
+  "second_${name}": 2
+in:
+  {$eval: "first_prize + second_prize"}
+
+context: {'name': "prize"}
+result:  3
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -156,7 +156,10 @@ def let(template, context):
 
     subcontext = context.copy()
     for k, v in template['$let'].items():
-        if not IDENTIFIER_RE.match(k):
+        if k in operators:
+            operator_values = renderValue(template['$let'], context)
+            subcontext = operator_values.copy()
+        elif not IDENTIFIER_RE.match(k):
             raise TemplateError('top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/')
         subcontext[k] = renderValue(v, context)
         if subcontext[k] == DeleteMarker:
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -156,14 +156,19 @@ def let(template, context):
 
     subcontext = context.copy()
     for k, v in template['$let'].items():
-        if k in operators:
-            operator_values = renderValue(template['$let'], context)
-            subcontext = operator_values.copy()
-        elif not IDENTIFIER_RE.match(k):
-            raise TemplateError('top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/')
-        subcontext[k] = renderValue(v, context)
-        if subcontext[k] == DeleteMarker:
-            subcontext[k] = ''
+        if IDENTIFIER_RE.match(k):
+            subcontext[k] = renderValue(v, context)
+            if subcontext[k] == DeleteMarker:
+                subcontext[k] = ''
+
+    initial_result = renderValue(template['$let'], context)
+    if not isinstance(initial_result, dict):
+        raise TemplateError("$let value must be an object")
+    for k, v in initial_result.items():
+        if not IDENTIFIER_RE.match(k):
+            raise TemplateError("top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/")
+        else:
+            subcontext[k] = v
     try:
         in_expression = template['in']
     except KeyError:
--- a/jsone/render.py
+++ b/jsone/render.py
@@ -155,12 +155,6 @@ def let(template, context):
         raise TemplateError("$let value must be an object")
 
     subcontext = context.copy()
-    for k, v in template['$let'].items():
-        if IDENTIFIER_RE.match(k):
-            subcontext[k] = renderValue(v, context)
-            if subcontext[k] == DeleteMarker:
-                subcontext[k] = ''
-
     initial_result = renderValue(template['$let'], context)
     if not isinstance(initial_result, dict):
         raise TemplateError("$let value must be an object")
--- a/specification.yml
+++ b/specification.yml
@@ -670,7 +670,7 @@ result:   35
 title:    let with evaluated context
 template: {$let: {$eval: '{"thirty": 10 + 20}'}, in: {$eval: "thirty+5"}}
 context:  {}
-error:    'TemplateError: top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/'
+result:    35
 ---
 title:    context overriding
 template: {$let: {x: 20}, in: {$eval: "x + y"}}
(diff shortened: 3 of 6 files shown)


owlishDeveloper:
I see the name says "WIP". Are you seeking any sort of specific feedback at this point? You don't say anything in the PR description.

richardgreg:
Hi @owlishDeveloper!

To be clear, is the goal to not restrict `$let` to have a key that corresponds with `/[a-zA-Z_][a-zA-Z0-9_]*/` before being evaluated?

owlishDeveloper:
@richardgreg  yes, the goal is to allow other things in there, that would **evaluate** to object (that would correspond to that regex)

richardgreg:
@owlishDeveloper Do you have any pointers about implementing the logic.

owlishDeveloper:
@richardgreg can you be more specific? What is the difficult? Where are you stuck? What have you tried?

richardgreg:
@owlishDeveloper thanks for the response; I have made some progress since I asked the question.

Assuming a solution is implemented, a statemet like the one below should return `35` instead of returning an error?
``` 
template: {$let: {$eval: '{"thirty": 10 + 20}'}, in: {$eval: "thirty+5"}}
```

richardgreg:
@owlishDeveloper this is only the JS implementation. I want feedback before writing the Python and GO code.

So, my solution first checks if the value of `$let` is an operator before moving on to the other sections of the function.  That is why
```
{$let: {$eval: '{"thirty": 10 + 20}'}, in: {$eval: "thirty+5"}}
```
also evaluates to `35`. This is the expected behaviour right?

richardgreg:
@owlishDeveloper I'm stuck. Do you have any suggestion(s) for me?

specification.yml
```diff
@@ -721,6 +721,11 @@ title:    checking $let with $if without else
 context:  {c: 'itm'}
 template: {$let: {a: {$if: 'c == "item"', then: 'value'}}, in: {a: '${a}'}}
 result:   {a: ''}
+---
+title:    let with a keyless nested operator
+template: {$let: {$if: something, then: {a: 10, b: 20}, else: {a: 20, b: 10}}, in: {$eval: 'a + b'}}
+context:  {'something'}
```
owlishDeveloper: Let's change the context to be something like this: `{something: 3}` and in the template: `$if: something` ---> `$if: 'something == 3'` to make the test case a bit clearer

richardgreg:
@owlishDeveloper My PR is ready for a review/merge. The bug-fix was right under my nose all along :laughing:

jsone/render.py
```diff
@@ -156,7 +156,10 @@ def let(template, context):
 
     subcontext = context.copy()
     for k, v in template['$let'].items():
-        if not IDENTIFIER_RE.match(k):
+        if k in operators:
+            operator_values = renderValue(template['$let'], context)
+            subcontext = operator_values.copy()
```
djmitche: This overwrites all existing values in `subcontext` :(

jsone/render.py
```diff
@@ -156,7 +156,10 @@ def let(template, context):
 
     subcontext = context.copy()
     for k, v in template['$let'].items():
-        if not IDENTIFIER_RE.match(k):
+        if k in operators:
```
djmitche: I don't think we want this conditional here.  It's possible to write a value that must be rendered but that does not contain an operator.  Rather than only render in some cases, we should render in *all* cases.

jsone/newsfragments/249.bugfix
```diff
@@ -0,0 +1 @@
+Fixed the issue of '$let' not evaluating if the key is an operator.
```
djmitche: From the perspective of a user of the language, this is not very helpful.  This PR is adding a new feature, so let's describe it that way.  Tell the user what they can do now, that they could not do before.  You can even use a small example.

jsone/render.py
richardgreg: Noted! I will get to that in a bit.

jsone/render.py
richardgreg: Okay!

richardgreg:
Hi @djmitche, nice of you to jump in. I've been awaiting a review! :)

I followed the steps you gave and made the implemetation. One test failed and I made a tweak to it, though I'm not 100% sure about it:

```yml
title:    checking $let with $if without else
context:  {c: 'itm', 'a':''}
template: {$let: {a: {$if: 'c == "item"', then: 'value'}}, in: {a: '${a}'}}
result:   {a: ''}
```

djmitche:
It's critical that we not change existing behavior, and that test represents existing behavior.  You'll need to make sure that test continues to pass without modification.

What you force-pushed does not implement the "new approach" I suggested.

src/index.js
```diff
@@ -145,15 +145,26 @@ operators.$let = (template, context) => {
...
-      throw new TemplateError('top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/');
-    }
-    variables[key] = render(template['$let'][key], context);
+    if (/^[a-zA-Z_][a-zA-Z0-9_]*$/.test(key)) {
+      variables[key] = render(template['$let'][key], context);
+    }  
     if (isObject(variables[key])) {
       variables[key] = '';
     }
   });
```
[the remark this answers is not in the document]
richardgreg: @djmitche getting rid of the loop causes the following statement to fail:
```yaml
template: {$let: {a: {$if: 'c == "item"', then: 'value'}}, in: {a: '${a}'}}
context:  {c: 'itm'}
result: unknown context value 'a'
```
Unless that is what we want.
Do you know how else I can we can grab the key wiithout looping?
I did clean things up a bit in the current implementation:
```js
  let variables = {};
  Object.keys(template['$let']).forEach(key => {
    if (/^[a-zA-Z_][a-zA-Z0-9_]*$/.test(key)) {
      variables[key] = render(template['$let'][key], context);
    }else{
      let initialResult = render(template['$let'], context);
      if (!isObject(initialResult)) {
        throw new TemplateError('$let value must be an object');
      }
      Object.keys(initialResult).forEach(key => {
        if (!/^[a-zA-Z_][a-zA-Z0-9_]*$/.test(key)) {
          throw new TemplateError('top level keys of $let must follow /[a-zA-Z_][a-zA-Z0-9_]*/');
        }else{
          variables[key] = initialResult[key];
        }
      });
      }
    if (isObject(variables[key])) {
      variables[key] = '';
    }
  });
```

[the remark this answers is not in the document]
djmitche: If I'm reading that correctly, you're saying that without the loop,
```yaml
title:    checking $let with $if without else 
context:  {c: 'itm'}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
template: {$let: {a: {$if: 'c == "item"', then: 'value'}}, in: {a: '${a}'}}
result:   {a: ''}
```
fails with "unknown context value 'a'".  That failure matches my expectations.  That's a case where `a` is, indeed, undefined (this is the DeleteMarker case I mentioned in my previous review).  In JSON-e we are careful to make things not exist when they do not exist, rather than to choose default values such as `''`.

I acknowledge that there is a bug in the current master where if the value of a $let assignment is an object, we set it to `''`:
https://github.com/taskcluster/json-e/blob/a3fdf13cd31d106448ff529021a1f77dac308f73/src/index.js#L152-L154

Since this PR is modifying that code, I'm happy to see that bug fixed at the same time.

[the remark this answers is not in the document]
djmitche: Here's a test case demonstrating the existing bug.  This fails on master but passes just fine with your new implementation, so is worth including in `specification.yml`:

```yaml
title:    $let with an object value                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
context:  {}
template: {$let: {a: {some: 'object'}}, in: {$eval: 'a'}}
result:   {some: 'object'}
```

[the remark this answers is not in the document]
richardgreg: Alright!

specification.yml
```diff
@@ -723,9 +723,39 @@ template: {$let: {x: [1, 2, 3]}, in: [$reverse: {$eval: "x"}]}
 result:   [[ 3, 2, 1 ]]
 ---
 title:    checking $let with $if without else
-context:  {c: 'itm'}
+context:  {c: 'item'}
 template: {$let: {a: {$if: 'c == "item"', then: 'value'}}, in: {a: '${a}'}}
-result:   {a: ''}
+result:   {a: value}
```
djmitche: Wait, this got changed around to not test the same thing anymore!  See my comment from a few hours ago -- with `{c: 'itm'}`, it should produce an error.

richardgreg: Simply giving the test case's result the value "unknown context value 'a'" does not pass when I run tests.

djmitche: Replace `result: ..` with `error: "unknown context value 'a'"` and it will check that the error occurs and matches that string.

richardgreg: `error: "unknown context value 'a'"` does not pass for any of the languages.
`error:    'InterpreterError at template.a: unknown context value a'` does, but for only Python and JS.

djmitche: OK, just use `error: true` then (which just checks that there's an error, without looking a the string)

richardgreg: Oops! `error: 'InterpreterError at template.a: unknown context value a'` passes for all languages. I was running the test on a similar but different branch from this current branch. Sorry for wasting both our time 😞. So I should just go with `error: 'InterpreterErrorr at...`?

specification.yml
```diff
@@ -725,7 +725,37 @@ result:   [[ 3, 2, 1 ]]
 title:    checking $let with $if without else
 context:  {c: 'itm'}
 template: {$let: {a: {$if: 'c == "item"', then: 'value'}}, in: {a: '${a}'}}
-result:   {a: ''}
+error:    'InterpreterError at template.a: unknown context value a'
```
djmitche: Note to later readers: this is technically a breaking change, but the behavior it changes (DeleteMarker resutling in a variable being set to an empty string) was introduced in #323 and hasn't been released yet, so there's no need to make this a major release.

Issue #158, opened 2017-07-13
`$json` should handle functions with an error -`{$json: {$eval: 'min'}}`


jonasfj:
`{$json: {$eval: 'min'}}` is the test case, it produces some weird error that isn't handled.
I think we want to explicitly check for this...

djmitche:
Ideally we don't have to recurse over the whole structure, but yes, we should handle this cleanly.  Isn't it acceptable to raise a JSON-e error?

jonasfj:
Yeah, some clean error is all we want..

viraj96:
@djmitche In the context of above test case, I presume the error originally is generated because no arguments are provided to `min`. As far as I have read the code, I see that `max` and `min` have `minArgs` defined as 1 and if `args.length` is less than `minArgs` then a `builtinError` is thown. So, now my question is what would `args.length` be in the above case because if it was a number then that `if` condition should have been triggered, right? Also, is there something that I am missing?

jonasfj:
> In the context of above test case, I presume the error originally is generated because no arguments are provided to min.

No that's not correct.

The value of the json-e expression `{$eval: 'min'}` is a function.
Now we don't allow json-e to evaluate to functions, or well, we somewhat do...

Example of something not allowed:
```yaml
$eval: 'min'  # not a legal json-e document, but it's a legal expression within another json-e document
```

Example of something allowed:
```yaml
$let:
  x: {$eval: 'min'}
in:
  test: {$eval: 'x(4,1)'}   # this evaluates to 1
```

So it's permitted to do `{$eval: 'min'}` which is useful in cases where you have a function that returns another function... like `{$eval: make_function("hello")("jonas")}`, if `make_function` takes a string and returns a function that also accepts a string then this is perfectly sane.. It also implies that functions can be values in json-e, but we only allow them as intermediate values.

Nicrii:
@owlishDeveloper I’d like to work on this issue

owlishDeveloper:
@Nicrii you're very welcome to it!

owlishDeveloper:
Fixed https://github.com/taskcluster/json-e/pull/330

owlishDeveloper:
The above fix did not cover all the test cases, so I re-open this. The test cases that should be covered:

```
# currently would return undefined but should throw a nice error:
template: {$json: {$eval: min}}
template: {$json: {$eval: fromNow}}
template: {$json: {$eval: '[min, max]'}}
[...]

PR #330, opened 2019-10-08
Issue 158

Nicrii:
`$json handle functions with an error -`{$json: {$eval: 'min'}}`

# Checklist

### DIFF
--- a/jsone.go
+++ b/jsone.go
@@ -50,6 +50,7 @@ func Render(template interface{}, context map[string]interface{}) (interface{}, 
 		}
 	}
 
+	// return result
 	return result, nil
 }
 
@@ -414,7 +415,7 @@ var operators = map[string]operator{
 		}
 		if !i.IsJSON(val) {
 			return nil, TemplateError{
-				Message:  "$json: can only stringify JSON types",
+				Message:  "$json can only stringify JSON types",
 				Template: template,
 			}
 		}
--- a/jsone.go
+++ b/jsone.go
@@ -407,7 +407,7 @@ var operators = map[string]operator{
 		}
 		if !i.IsJSON(val) {
 			return nil, TemplateError{
-				Message:  "$json can only stringify JSON types",
+				Message:  "$json: function in template must have arguments",
 				Template: template,
 			}
 		}
--- a/jsone.go
+++ b/jsone.go
@@ -414,7 +414,7 @@ var operators = map[string]operator{
 		}
 		if !i.IsJSON(val) {
 			return nil, TemplateError{
-				Message:  "$json: function in template must have arguments",
+				Message:  "$json: can only stringify JSON types",
 				Template: template,
 			}
 		}
--- a/jsone.go
+++ b/jsone.go
@@ -45,7 +45,7 @@ func Render(template interface{}, context map[string]interface{}) (interface{}, 
 
 	if result != nil && reflect.TypeOf(result).Kind() == reflect.Ptr {
 		return nil, TemplateError{
-			Message:  "$eval: function dont get any arguments in template",
+			Message:  "$eval: function doesn't get any arguments in template",
 			Template: template,
 		}
 	}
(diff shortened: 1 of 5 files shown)


owlishDeveloper:
My question here is: are we sure the problem is in the fact that `min` doesn't have arguments? This question has actually been already raised in the issue. [As per Jonas's reply,](https://github.com/taskcluster/json-e/issues/158#issuecomment-334585277) in certain cases it's completely fine that `min` doesn't have arguments....

The way I understand the issue is this: the code in the first message on the issue produces this error: `unacceptable kind of an object to dump [object Undefined]`, and `$eval: 'min'` produces `unacceptable kind of an object to dump [object Function]`. These errors are thrown by js-yaml library. We want to handle this error and provide our own error message instead.

Nicrii:
By "min must have arguments" I meant that min has no arguments in the whole context, and not just where it is called, which seems to fit the situation discussed in reply.
I will reconsider my approach to the problem. This error should be handled not only in $json?I'm little confused

Nicrii:
Now I see my mistake. I'll fix it as soon as possible. Thanks for explanation

Nicrii:
But I'm thinking about naming this error. Isn't the problem in reply that the function was called, but in template was't argument's for it? In different example there are were argument's, but in different place in template, so everything work right

owlishDeveloper:
I discussed the issue with my colleague, and I am back to you with some refined requirements! (I don't think the issue is phrased in a very clear manner; there's also a mention of some error which is probably not there anymore in the newer version of node)

So the JSON language doesn't process functions. If you give it a function, it would just return `undefined` (at least in modern versions of Node). The confusion about arguments comes from the fact that the function can be either called or not. If it's called (like in `$eval: 'min(2, 3)`), we are basically evaluating the return value of the function, which JSON is ok with. If we don't call the function (like in `$eval: min`) we are evaluating the function itself which JSON doesn't allow.

Now, it's all good, but in our language, JSON-e, we don't want to just return `undefined` - we want to throw an error (such as `$json cannot convert non-JSON-able values such as functions` or similar message).

Some examples of test cases:

```
# currently would return undefined but should throw a nice error:
template: {$json: {$eval: min}}
template: {$json: {$eval: fromNow}}
template: {$json: {$eval: '[min, max]'}}
template: {$json: [{$eval: 'min'}]}
template: {$json: {maximum: {$eval: 'max'}}}
template: {$json: [{minimum: {$eval: 'min'}, {maximum: {$eval: 'max'}}]}

# should return strings containing return values of the functions
template: {$json: [min, max]}
template: {$json: {$eval: "fromNow('1 day')"}}
template: {$json: [{$eval: 'min(2, 3)'}]}
template: {$json: {minimum: {$eval: 'min(1, 2)'}}}
```

Let me know if this makes more sense and if you have more questions!

jsone/render.py
```diff
@@ -142,6 +142,8 @@ def ifConstruct(template, context):
 def jsonConstruct(template, context):
     checkUndefinedProperties(template, ['\$json'])
     value = renderValue(template['$json'], context)
+    if callable(value):
+        raise TemplateError('$json returns undefined property')
```
owlishDeveloper: Let's add some info to the message: `$json returns undefined property because functions are not allowed in JSON`

jsone.go
```diff
@@ -37,11 +38,18 @@ func Render(template interface{}, context map[string]interface{}) (interface{},
...
 
-	// Handle deleteMarker
+	//Handle deleteMarker
 	if result == deleteMarker {
 		result = nil
 	}
 
+	if result != nil && reflect.TypeOf(result).Kind() == reflect.Ptr {
+		return nil, TemplateError{
+			Message:  "$eval: function dont get any arguments in template",
```
owlishDeveloper: typo: `dont` --> `doesn't`

and let's make this message consistent across implementations!

owlishDeveloper:
Thank you for the fix!

PR #340, opened 2019-12-16
Adding AST and making parser

Nicrii:
I'm trying to use tokenizer from Prattparser in my parser in python implementation(as I already did in js), but I have problems with imports modules. I found one decision, but it's disturb  the existing code.


specification.yml
```diff
@@ -17,5048 +17,8 @@
 ---
 section:  Initial Minimal Subset
 ---
-title:    Identity
```
owlishDeveloper: Is this part of the existing code disturbance? the specs should not go anywhere, because not only they document the library, but also they help testing the functionality. Our task is to get them working with the new way of parsing.

Nicrii: No, it was not. 
This was for my purposes, it should not be here, i've added this to my commits by mistake, sorry

src/newparser.js
```diff
@@ -0,0 +1,45 @@
...
+            node = new UnaryOp(token, this.factor());
+        } else if (token.kind == "!") {
+            this.eat("!");
+            node = new UnaryOp(token, this.factor());
+        } else if (token.kind == "number") {
+            this.eat("number");
+            node = new Term(token);
+        }
+
+        return node;
```
owlishDeveloper: I see this function returns a node. Do you have a plan for where the nodes will be assembled into a tree?

Nicrii: Yes, there will be function for tree traversal.

src/AST.js
```diff
@@ -0,0 +1,43 @@
+class ASTnode {
+    constructor(token) {
+        this.token = token;
```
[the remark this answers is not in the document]
Nicrii: There are possibility that our tree won't be binary? I didn't see any not binary or unary operations. What I missed?

jsone/prattparser.py
```diff
@@ -2,8 +2,8 @@
 
 import re
 from collections import namedtuple
-from .shared import TemplateError
-from .six import with_metaclass, viewitems
```
Nicrii: This breaks existing tests, but I didn't find other way to resolve problems with imports in my AST_test.py

src/AST.js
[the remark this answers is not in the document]
Nicrii: I found some cases, the question is cleared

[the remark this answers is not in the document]
owlishDeveloper: Oh, I meant binary tree - as in, a tree where each node has no more than two children (apologies for not being clear!)

src/AST.js
```diff
@@ -0,0 +1,35 @@
+class ASTNode {
```
owlishDeveloper: I have to admit that I was wrong about the name of this class. This should be a _tree_, not a node, and its children are subtrees, and not just nodes. Feel free to rename this if you'd like

src/newparser.js
```diff
@@ -0,0 +1,96 @@
+const {ASTNode, UnaryOp, BinOp} = require("../src/AST");
+const Tokenizer = require("../src/tokenizer");
+
+class Parser {
+    constructor(tokenizer, source, offset = 0) {
+        this._source = source;
+        this._tokenizer = tokenizer;
+        this.current_token = this._tokenizer.next(this._source, offset);
```
owlishDeveloper: Why are we calling `tokenizer.next` in the constructor?

Nicrii: To take first token

src/newinterpreter.js
```diff
@@ -0,0 +1,171 @@
...
+            case ("+"):
+                return +this.visit(node.expr);
+            case ("-"):
+                return -this.visit(node.expr);
+            case ("!"):
+                return !this.visit(node.expr);
+        }
+    }
+
+    visit_BinOp(node) {
```
owlishDeveloper: Many of these will require type checking and the error handling, I think

src/newparser.js
```diff
@@ -0,0 +1,253 @@
...
+            this.eat(token.kind);
+            node = new BinOp(token, node, next.call(this));
+        }
+
+        return node
+    }
+
+    parse() {
+        //    expr : logicalAnd (OR logicalAnd)*
+        return this.simpleRule.call(this, this.logicalAnd, ["||"])
```
owlishDeveloper: This function is called `parse` but it seems it handles logical OR operator.... Probably should be called `logicalOr` instead?

owlishDeveloper: Another question about this function and the function below. Do you plan any other code in these? If not, it might be a good idea to get rid of the `simpleRule ` function and transfer the code in these functions. I realize I am asking you to duplicate some boilerplate code, but I don't like that all these are coupled, and I don't like the readability of a function that does nothing but calls another function.

You don't have to do the refactor at this stage tho. We can revisit this at the end of the work. Obviously feel free to refactor at any point of you find the current organization inconvenient.

src/newparser.js
```diff
@@ -0,0 +1,253 @@
...
+        //    multiplication : exponentiation (MUL | DIV exponentiation)*
+        return this.simpleRule.call(this, this.exponentiation, ["*", "/"])
+    }
+
+    exponentiation() {
+        //    exponentiation : factor (EXP factor)*
+        return this.simpleRule.call(this, this.factor, ["**"])
+    }
+
+    factor() {
```
owlishDeveloper: This actually seems to be a `parse` function? As far as I can tell it parses tokens into the AST, doesn't it?

src/newparser.js
```diff
@@ -0,0 +1,253 @@
...
+
+    multiplication() {
+        //    multiplication : exponentiation (MUL | DIV exponentiation)*
+        return this.simpleRule.call(this, this.exponentiation, ["*", "/"])
+    }
+
+    exponentiation() {
+        //    exponentiation : factor (EXP factor)*
+        return this.simpleRule.call(this, this.factor, ["**"])
+    }
```
owlishDeveloper: Is there an alternative to organize logic into a chain of little functions that call each other? Maybe a recursion, or some other kind of a loop? This code might end up being difficult to read and understand, because people will have to follow little functions like breadcrumbs.

src/newparser.js
Nicrii: No, it's just part of parser, it represents one of grammar rules

src/newparser.js
Nicrii: I can return previous version of code, where was no `simpleRule` and was duplicating code
About `parse`, if we will change back this name to the name of the grammar rule, would it be better to return [this](https://github.com/taskcluster/json-e/pull/340#discussion_r358560220)

src/newparser.js
Nicrii: I'll think about different code organization

src/parser.js
```diff
@@ -0,0 +1,321 @@
...
+
+        for (; token !== null && token.kind == "||"; token = this.current_token) {
+            this.eat(token.kind);
+            node = new BinOp(token, node, this.logicalAnd());
+        }
+
+        return node
+    }
+
+    logicalAnd() {
```
owlishDeveloper: I see all these functions call each other in a chain - seems like that's your way of expressing recursion. I wonder if we could come up with a shorter way to do this?

jsone/AST.py
```diff
@@ -0,0 +1,66 @@
+class ASTNode(object):
+    def __init__(self, token):
+        self.token = token
+
+    def __eq__(self, other):
```
owlishDeveloper: 🤔 This is for the tests? I don't entirely enthusiastic about code written exclusively for tests.... Is it possible to implement this outside of this class? In the test module somewhere?

jsone/parser.py
```diff
@@ -0,0 +1,257 @@
...
+                args.append(node)
+            self.takeToken(")")
+
+        node = Builtin(token, args)
+        token = self.current_token
+
+        while token is not None and token.kind == ".":
+            self.takeToken(".")
+            right = ASTNode(self.current_token)
+            self.takeToken(self.current_token.kind)
```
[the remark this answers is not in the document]
Nicrii: I changed this syntax error, from
`'SyntaxError: Found {, expected !, (, +, -, [, false, identifier, null, number, string, true, {'`
to 
`'SyntaxError: Found {, expected identifier'`
now it's match your example, but I had to change this:
`title: missing right square bracket`
`context:  {bar: "mystring"}`
`template: {foo: '${bar[2:}'}`
`error:   "SyntaxError at template.foo: Found }, expected !, (, +, -, [, false, identifier, null, number, string, true, {"`
to
`error:   "SyntaxError at template.foo: Found }, expected ]"` 
@owlishDeveloper  [didn't like this version of error name](https://github.com/taskcluster/json-e/pull/340#discussion_r377970864)
I'm not sure is there good way to change first error name and not to change second error name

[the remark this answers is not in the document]
djmitche: I agree with @owlishDeveloper that it is a bug that these lists of expected tokens are changing.  For example, in 
```yaml
template: {$eval: 'key.{'}                                                                                                                                                                                                                                                                                                     
```
only an identifier is allowed there, as indicated by the original error message.  As it had been written when I reviewed yesterday, `key.true` was also allowed (although it failed during `visit`), which is a real bug and not just a different error message.  So I'm happy to see that you've fixed that!

However, the newly changed error also suggests a bug.  At the spot where `}` occurs in
```yaml
template: {foo: '${bar[2:}'}
```
more than just `]` would be allowed -- in fact, any expression is allowed.  So the original error message is correct, and I would expect it to come from the `parse` method.

jsone/interpreter.py
```diff
@@ -159,19 +159,19 @@ def visit_ValueAccess(self, node):
...
         args = []
         try:
-            builtin = self.context[node.token.value]
+            contextValue = self.context[node.token.value]
         except KeyError:
             raise InterpreterError(
                 'unknown context value {}'.format(node.token.value))
-        if callable(builtin):
+        if callable(contextValue):
             if node.args is not None:
```
djmitche: It looks like this is the implementation of function invocation, embedded in a ContextValue.  But, the case where we try to call something that is not callable is not treated as an error.  So something like
```yaml
context: {function: "hi"}
template: {$eval: "function(10)"}
error: true
```
would fail (`AssertionError: expected exception`).  It would be good to add that as a test case.  I see that it gives a particularly weird error in `master` on Python, and no error on JavaScript or Golang, so this might be a corner case we didn't address very well in the past!

That function calls are bundled into ContextValue implies an assumption that a function call always has the form `IDENTIFIER(ARG, ..)`.  But that's not necessarily the case:

```yaml
context: {}
template: {$let: {functions: [{$eval: min}, {$eval: max}]}, in: {$eval: 'functions[1](2, 3)'}}
result: 3
```

It might make sense to have two different AST node types, ContextValue and FunctionCall?

jsone/parser.py
```diff
@@ -36,164 +37,178 @@ def takeToken(self, *kinds):
...
-            node = self.object()
+            node = self.parse_object()
+            node = self.parse_dot_operation(node)
         elif token.kind == "identifier":
-            node = self.builtins()
-            node = self.valueAccess(node)
+            node = self.parse_context_value()
+            node = self.parse_dot_operation(node)
+            node = self.parse_value_access(node)
+            node = self.parse_dot_operation(node)
```
djmitche: Hm, why does `parse_dot_operation` appear here twice?  It looks like it is here to make a test case with `x.y.z` pass, but would still fail `w.x.y.z`?  But, that's wrong because this test case passes on the AST branch:
```yaml
title: dots on dots
context: {a: {b: {c: {d: {e: {f: 10}}}}}}
template: {$eval: 'a.b.c.d.e.f'}
result: 10
```
so I'm missing something here :)

I wonder if it makes sense to parse `.` as a binary operation like the others, just one with higher precedence than all of the others (so that `x.y**2` parses as `BinOp('**', BinOp('.', 'x', 'y'), 2)`).

jsone/parser.py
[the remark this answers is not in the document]
Nicrii: I've returned original error for this particular case, but I have difficulty determining the general case / other places causing errors of this type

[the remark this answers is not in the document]
djmitche: I think that aligning the grammar the parser implements with the existing grammar will help.  As a double-check and to find any other areas of difference, you could invent a few test cases with invalid inputs and add them to the specification.

jsone/parser.py
```diff
@@ -64,26 +66,27 @@ def parse(self, level=0):
 
         return node
 
-    def parse_property_access(self):
-        """  propertyAccess : unit (accessWithBrackets (functionCall)? | DOT id)* """
+    def parse_property_access_or_func(self):
+        """  propertyAccessOrFunc : unit (accessWithBrackets | DOT id | functionCall)* """
```
djmitche: (excuse me while I think out loud about the expression grammar..)

The PrattParser had precedence set like this:
https://github.com/taskcluster/json-e/blob/bce0ed6f9bdaa2a863c768c6b966922fe19105a4/jsone/interpreter.py#L57-L58
meaning that property access has lower precedence than function evaluation.  In the PrattParser, precedence comes into play when a binary operation recursively parses with a higher precedence level (so `+` parses with level `*`).  But none of the defintiions of `.`, `[`, and `(` recursively parse their right-hand-side like that.  `[` and `(` both parse until a terminating `]` or `)`, while `.` requires an identifier on its right-hand side, and not an expression.  So the difference in precedence doesn't really matter, and we could have written `["[", "(", "."]` in the PrattParser precedence map.

Which is equivalent to what you've done here.

Thanks for listening to me figure this out :)  If I got anything wrong, please let me know!

jsone/parser.py
```diff
@@ -0,0 +1,252 @@
...
+
+        return node
+
+
+def parse_string(string):
+    return string[1:-1]
+
+
+class Tokenizer(object):
+    def __init__(cls):
```
djmitche: `__init__` gets an instance of the class as its argument.  Calling it `cls` instead of `self` doesn't change that :)

I suspect you're leaning toward the metaclass approach that the old PrattParser class used.  I think that was probably more clever than necessary.  I think the only thing you'd need to change (aside from this confusing variable name) is to call `tokenizer()` once at module level in `interpreter.py` and use the resulting value from `render`, instead of calling `tokenizer()` at every call to `render`.

jsone.go
```diff
@@ -13,7 +13,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	i "github.com/taskcluster/json-e/interpreter"
+	i "./interpreter"
```
djmitche: This change, and a few like it, seems to be breaking the go test run in CI, and didn't work for me either.  I'm not a big Gofer, but I've never seen this syntax before.  Is it something new with Go1.14?

djmitche: ("The fix" is just reverting this change and a few other places with the same kind of thing)

owlishDeveloper:
@djmitche thank you so much for your thorough review!

owlishDeveloper:
It seems to me there used to be ` jsone/newsfragments/244.bugfix` in here? I don't see it now; let's add it! This will be the changelog for the new release. Let's mention there that we add support for the short-circuiting of the boolean logic operators `||` and `&&`, and the fact that json-e now uses AST.

PR #344, opened 2020-02-03
add a new $fromNow test and rename to distinguish from builtin

djmitche:
Adds a test for `$fromNow` using a `now` value defined in a `$let`.  Also renames tests for `$fromNow` to include the `$` to avoid ambiguity.

# Checklist

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -466,119 +466,124 @@ template: {$flattenDeep: [[1, [2, 3]], [4], 5], foo: 'bar'}
 error: 'TemplateError: $flattenDeep has undefined properties: foo'
 ################################################################################
 ---
-section:  fromNow
+section:  $fromNow
 ---
-title:    fromNow 1 hour
+title:    $fromNow 1 hour
 context:  {}
 template: {$fromNow: '1 hour'}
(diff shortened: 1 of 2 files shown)


specification.yml
```diff
@@ -466,119 +466,124 @@ template: {$flattenDeep: [[1, [2, 3]], [4], 5], foo: 'bar'}
...
 template: {$fromNow: 13}
 error:    'TemplateError: $fromNow expects a string'
 ---
-title:    fromNow with eval
+title:    $fromNow with eval
 context:  {ttl: 24}
 template: {$fromNow: {$eval: 'str(ttl) + " hours"'}}
 result:   '2017-01-20T16:27:20.974Z'
 ---
-title:    fromNow with reference
```
owlishDeveloper: why are we deleting this test-case? It seems to be different from the ones you're adding

djmitche: It's not deleted - it's the same test as the second one (line 581-584) but I changed the name `now` to `when` to make sure that `$fromNow` is using the `from` parameter and not defaulting to `now`.

owlishDeveloper: oops didn't see :) ok, ty!

Issue #343, opened 2020-02-03
fromNow builtin behaves differently in different languages


djmitche:
```yaml
title:    fromNow with redefined `now`
context:  {}
template: {$let: {now: '2017-01-01T01:00:00.123Z'}, in: {$eval: "fromNow('1 day')"}}
result:   '2017-01-02T01:00:00.123Z'
```

fails on JS and Python and succeeds on Go.  This is similar to #342 and will probably have a similar solution.  @Nicrii if it's easy to fix both at once, that'd be cool, otherwise I'm happy to fix this up following the approach you choose for #342.

owlishDeveloper:
I have a question that is similar to the one I left in #342 - can you tell us more about the practical problem you are trying to solve? I am asking because the `fromNow` built-in accepts 2 arguments: the distance (`str` argument - in this case it's `1 day`) and the optional starting point (`reference` argument) - if you do

```yml
title:    fromNow with redefined `now`
context:  {}
template: {$eval: "fromNow('1 day', '2017-01-01T01:00:00.123Z')"}
result:   '2017-01-02T01:00:00.123Z'
```
as opposed to defining a variable, the test case will pass. So I am curious to learn more about the context (pun not intended)

djmitche:
@Nicrii fixed this in #345

Issue #342, opened 2020-02-03
defined behaves differently in different implementations


djmitche:
The following test case:
```yaml
---
title: 'defined with empty variable from a let expression'
context: {}
template: {$let: {var: 10}, in: {$if: "defined('var')", then: 't', else: 'f'}}
result: t
```
fails on JS and Python, but succeeds on Golang.  The test case demonstrates the expected behavior, though: a variable defined with `$let` should be visible to `defined`. I think the issue is that in the JS and Python implementations, the `context` variable that is being referred to in the implementation of `defined` is the original, top-level `context` to which the builtins were added, rather than the `context` that's active at the time the expression is evaluated.

Since this behavior differs between implementations, it will be important to fix this before we make another release.

@Nicrii is there any chance you could have a look?

owlishDeveloper:
Can you tell us more about the practical problem that's behind this test case? I'm asking because one way of going about checking whether a variable is defined would be to use the concept of `truthy`/`falsy`, as in javascript language:

```javascript
if (var) return 't'; else return 'f'
```

as opposed to using `defined` built-in.... but maybe I am missing something?

owlishDeveloper:
ok I went back down the issues and PRs, makes sense now!

djmitche:
@Nicrii fixed this in #345.

PR #345, opened 2020-02-04
$let now update builtins in the context

Nicrii:
Issue #342 #343

# Checklist


src/index.js
```diff
@@ -157,7 +157,7 @@ operators.$let = (template, context) => {
     }
   });
 
-  var child_context = Object.assign(context, variables);
+  let child_context = Object.assign(context, addBuiltins(Object.assign({}, variables)));
```
djmitche: There's another bug here, that already existed: `Object.assign` modifies its first argument.  You've done the right thing in the code you've added (`Object.assign({}, ..`), but the existing call to `Object.assign` should be fixed, too.  Otherwise this call will modify the passed-in context.  I'm not sure how to write a test to demonstrate this error, but I think it should still be fixed :)

Nicrii:
How about this?

djmitche:
Ready for another look?

Nicrii:
Yes

test/test_misc.py
```diff
@@ -8,7 +8,7 @@
 
 
 def test_custom_builtin():
-    def my_builtin(x, y):
+    def my_builtin(ctx, x, y):
```
djmitche: Uh-oh!  This, and the similar change in `test/misc_test.js`, are breaking changes as they change the signature that custom built-ins must use.  It is possible to make a major release and announce this difference, but it would be better to avoid breaking compatibility!

I see that in the Python and JS implementations, you've added special cases for `defined` and `fromNow`, but no such special-casing is applied in Python for functions that are not variadic and do not have argument tests.  I think that's why this test case fails without this change.

The approach you've taken is very similar to that in the Go implementation -- but the Go implementation uses `i.WrapFunctionWithContext` and `i.WrapFunction` to tag functions as needing, or not needing, context.  This moves the information about whether a function expects context to where that function is defined, rather than farther away in the function-invocation code.  Could you do something similar in Python and JS?  In both languages, you can add properties to function objects, so perhaps a `needs_context` property?

I think this is really close -- we just need to get to a state where these two test cases aren't changed, and hopefully a minor refactor to make it clearer in Python and JS which functions get a context argument.

Nicrii: Test fails because of context in function call, in js it didn't fail as context is last argument in call, so js just ignore it. if you put context to first place there will be the same error. Also, I didn't find tests for custom builtins in go. Is there place in go code where custom builtins are wrapping?

djmitche: JS's relaxed attitude toward function arguments was an advantage here :)

That's a good question about Go custom functions.  But since this PR isn't changing the Go implementation, let's not worry about it right now.  That would be a good issue to file, if you'd like (to add some tests).

djmitche:
@Nicrii will you get a chance to wrap this up?

djmitche:
Or are you waiting for my review?  You should see a little button next to my name that you can click to request that:
image

jsone/interpreter.py
```diff
@@ -189,7 +189,10 @@ def function_call(self, left, token, pc):
         if not callable(left):
             raise TemplateError('function call', 'callable')
         args = parseList(pc, ',', ')')
-        return left(*args)
+        if 'build' in left.__globals__:
```
djmitche: This had be scratching my head for a while -- a long time!

My best guess is, this is a way to guess whether the function is a builtin or not, taking advantage of the fact that there's a global in the `jsone.builtins` module named `build` -- and implicitly assuming that no user code would ever have a global variable with that name.  Imagine how much time a user would spend debugging an error where an unrelated change that introduces a global named `build` causes their JSON-e renderings to fail in mysterious ways!

The principle is sound, though: we need a way to reliably distinguish a builtin (which always gets `context`) from a user-supplied function. To do so reliably, it would be good to use something that explicitly names `jsone`.  One option might be to add a `_jsone_builtin` property to every function constructed with the `builtin` decorator, and then use `hasattr(left, '_jsone_builtin')` here.

This approach is *very* close to your approach, but less surprising for users.

src/interpreter.js
```diff
@@ -262,7 +262,11 @@ infixRules['.'] = (left, token, ctx) => {
 
 infixRules['('] =  (left, token, ctx) => {
   if (isFunction(left)) {
-    return left.apply(null, parseList(ctx, ',', ')'));
+    let args = parseList(ctx, ',', ')');
+    if (left.name == "") {
```
djmitche: Similar notes here to that for the Python approach.  It's not unlikely that a user would supply a function to which they'd attached a `name` property.  It's more unlikely that they'd supply one with a `_jsone_builtin` property (and less surprising to them when doing so causes errors).

jsone/builtins.py
```diff
@@ -23,23 +23,32 @@ def invoke(*args):
...
                         if not variadic(arg):
                             raise BuiltinError('invalid arguments to builtin: {}'.format(name))
+                    if needs_context is True:
+                        return fn(context, *args)
                     return fn(*args)
 
             elif argument_tests:
-                def invoke(*args):
+                def invoke(context, *args):
+                    _jsone_builtin = True
```
djmitche: This bit is unnecessary (and an unused local variable) -- the bit on line 48 is the one that sets the attribute.  I'll push a commit to remove this :)

Issue #354, opened 2020-03-19
Functions call does not work in v4.x


the-question:
Reproduction in Playground:  
https://json-e.js.org/#Playground/%7B%7D&%7B%24eval%3A%20'max(3%2C%202)'%7D

djmitche:
```yarn
template: {$eval: 'max(3, 2)'}
context: {}
```

the-question:
The problem seems to come from https://github.com/taskcluster/json-e/blob/master/src/interpreter.js#L12 where `e.constructor.name` returns an empty string

djmitche:
I see this working on master, actually (see test case PR).  It appears it's not working in the playground perhaps because the playground is running a different version?  Do you see this on a checkout of v4.0.1 or master?

the-question:
I have this problem on v4.0.1, I did not test it against master

djmitche:
Hm, I've been unable to reproduce.  Can you show a code snippet that reproduces?

the-question:
On the website's Playground, I get an error in the result with:
```
template: {$eval: 'max(3, 2)'}
context: {}
```

I don't have the error in NodeJS. It also seems I don't have the error when webpack builds it in dev mode. But I do have the error when webpack build in production mode.
I made a really dirty CodeSandbox: https://codesandbox.io/s/goofy-shape-pi3q0
To reproduce, you can:
- export the ZIP
- `npm install`
- `npm run build` to build using webpack in production mode
- open the resulting `build/index.html` in the browser. In the console, you will see the error: `TypeError: this[("visit_" + e.constructor.name)] is not a function`

My apologies for the dirty reproduce, I tried to make something ASAP.
Thanks a lot for your time and your reactivity!

the-question:
@djmitche I found out that Webpack get's rid of constructor's name in production mode. I created #356  to replace usage of `constructor.name` with a manually set property `constructorName`. It's slightly less elegant than using the constructor's name but at least it works consistently. Could you review this PR please?

the-question:
@djmitche Anything I could do to help move this forward? This is currently a blocker for me to upgrade to 4.x

PR #356, opened 2020-03-19
354 fix constructor name

the-question:
When using Webpack in production mode, constructor's name are optimized away. This leads to a TypeError in `src/interpreter.js`'s `visit` function.
# Checklist

### DIFF
--- a/jsone/newsfragments/354.bugfix
+++ b/jsone/newsfragments/354.bugfix
@@ -1,0 +1,1 @@
+Fix error (constructor.name) when using with Webpack in production mode
--- a/src/AST.js
+++ b/src/AST.js
@@ -1,6 +1,7 @@
 class ASTNode {
     constructor(token) {
         this.token = token;
+        this.constructorName = 'ASTNode';
     }
 }
 
@@ -9,6 +10,7 @@ let Primitive = ASTNode;
 class BinOp extends ASTNode {
     constructor(token, left, right) {
         super(token);
+        this.constructorName = 'BinOp';
         this.left = left;
         this.right = right;
     }
@@ -17,6 +19,7 @@ class BinOp extends ASTNode {
 class UnaryOp extends ASTNode {
     constructor(token, expr) {
         super(token);
+        this.constructorName = 'UnaryOp';
         this.expr = expr;
     }
 }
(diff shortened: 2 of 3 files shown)


owlishDeveloper:
@the-question can you tell us a bit more about the error and how it occurs?

the-question:
@owlishDeveloper When Webpack runs in production mode, it makes some optimizations/[minifications](https://webpack.js.org/guides/production/#minification) to reduce the bundle size (including uglifying).
Specifically, the [Terser Plugin](https://webpack.js.org/plugins/terser-webpack-plugin/) does the minifying part. Both compression and mangling (replacing variable names with shorter ones) is performed. There is a `keep_classnames` setting both in the compression and mangling part, and a general one which can control these two. By default, Terser does not keep classnames, which means it will both attempt to mangle them, but more importantly compress (aka remove) them.

In #354 I linked [an issue](https://github.com/webpack/webpack/issues/6841) from Webpack's git itself. It is generally [not recommended](https://dev.to/gaurangdhorda/constructor-name-is-not-working-in-production-3ik4) to use `constructor.name` for production due to multiple tools removing it when optimizing code for production. Counterintuitively, as ES6 standardized usage of `constructor.name`, more and more are having problem with it because of transpilers/bundlers/you name it, who remove it (e.g: [ember](https://discuss.emberjs.com/t/constructor-name-behaves-differently-in-dev-and-prod-builds-for-models-defined-with-the-es6-class-syntax/15572), [general discussion about the problem](https://medium.com/@eransakal/fix-javascript-issues-with-function-name-7079742f1a05), related issues [[1](https://github.com/terser/terser/issues/468), [2](https://github.com/terser/terser/issues/118)] in Terser's git).

djmitche:
@owlishDeveloper can you have a look at the PR given that answer?

owlishDeveloper:
> In #354 I linked [an issue](https://github.com/webpack/webpack/issues/6841) from Webpack's git itself

I am noticing that the issue is closed and the author got a recommendation on how to handle that - have you tried following [the advice in the issue](https://github.com/webpack/webpack/issues/6841#issuecomment-375713086)?

I personally don't see any harm in merging this, however.

the-question:
The advice in the issue would probably work, except that the module's code would not be optimized (minimized) in that case.

Thank you very much for approving these changes @owlishDeveloper, it will make json-e more friendly (just works) with bundlers :tada:

Issue #257, opened 2018-09-18
$matchOne, a variant of $match


djmitche:
I wanted to do this:

```
priority:                                                                                                                                                                                                                                                     
  $match:
    "tasks_for == 'decision'": low 
    "tasks_for == 'cron'": very-low
    "tasks_for == 'action'": lowest
```
https://hg.mozilla.org/try/rev/d39a2a66e72d

But the result was `{priority: ['low']`}` -- the array is unnecessary and undesired.  But it's there since in principle more than one of the options might have matched.

So, this issue proposes an operator similar to `$match`, but where exactly one option must match or the operator fails; and the result does not have an array.  An example of usage would be changing `$match` to `$matchOnce` in the above and getting the result `{priority: low}`.

Callek:
Slight preference for $matchOne with the added caveat that I want it deterministic so I feel if more than one match we should fail out when it's used

djmitche:
Thanks, and agreed regarding determinism!  Which reminds me, #258.

owlishDeveloper:
isn't that a `$switch`?

djmitche:
That is probably a better name :)

olsenme:
I'd like to work on this issue

owlishDeveloper:
@olsenme  have you sent me the screenshots?

stuartelimu:
Can I work on this issue?

owlishDeveloper:
@stuartelimu it's all yours!

stuartelimu:
thank you, on it now

stuartelimu:
thank you so much

richardgreg:
@stuartelimu are you still working on this issue?

stuartelimu:
Yes, I submitted a draft PR.  @owlishDeveloper is reviewing it

owlishDeveloper:
an update: this issue is still open and relevant.

noahlaux:
alternatively you could use a merge to achieve the same thing

```yaml
priority:
  $merge:                                                                                                                                                                                                                                                     
    $match:
      "tasks_for == 'decision'": low 
      "tasks_for == 'cron'": very-low
      "tasks_for == 'action'": lowest
```

djmitche:
That's true!  But @Callek's point about determinism is a good one.  It'd be tempting to write things like 
```
priority:
  $merge:                                                                                                                                                                                                                                                     
    $match:
      'x == 10': ten 
      'x > 5': five
      'x <= 5': small
```
[...]

PR #358, opened 2020-05-29
Feature switch

noahlaux:
This change introduces a new `$switch` operator which behaves like a combination of the `$if` and `$match` operator for more complex boolean logic. It gets an object, in which every key is a string expression(s), where only one must evaluate to true and the remaining to false based on the context. The result will be the value corresponding to the key that were evaluated to true.

If there are no matches, the result is either null or if used within an object or array, omitted from the parent object.

The idea was originally discussed in https://github.com/taskcluster/json-e/issues/257

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,13 +396,13 @@ result:   [3, 6, 9]
 ```
 The array or object is the value of the `$map` property, and the expression to evaluate
 is given by `each(var[,key|index])` where `var` is the name of the variable containing each
-element and `key|index` is either the object key or array index of the value. In the case of 
-iterating over an object and no `key|index` var name is given, `var` will be an object with 
+element and `key|index` is either the object key or array index of the value. In the case of
+iterating over an object and no `key|index` var name is given, `var` will be an object with
 two keys: `key` and `val`. These keys correspond to a key in the object and its corresponding value.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -447,6 +447,54 @@ context: {x: 10}
 result: []
 ```
 
+### `$switch`
+
+The `$switch` operator behaves like a combination of the `$if` and
+`$match` operator for more complex boolean logic. It gets an object,
+in which every key is a string expression(s), where only *one* must
```
djmitche, line 454:
```suggestion
in which every key is a string expression(s), where at most *one* must
```

jsone/newsfragments/257.feature
```diff
@@ -0,0 +1,4 @@
+Introduces $switch operator, which behaves like a combination of the $if and $match operator for
+more complex boolean logic. It gets an object, in which every key is a string expression(s), where
+only one must evaluate to true and the remaining to false based on the context. The result will be
```
djmitche, line 3:
```suggestion
at most one must evaluate to true and the remaining to false based on the context. The result will be
```

noahlaux:
Happy you liked it and sorry for my poor grammar :)

djmitche:
Published in 4.1.0 -- https://json-e.js.org/#Language/-switch-

Issue #368, opened 2020-08-18
New feature: `split`


noahlaux:
# Split

This issue suggest to create a builtin `split` functionality as available in many languages, and is not possible using any other (or combination of) exciting builtins.

Splitting on strings with a delimiter is the most obvious, but this suggestion also give a take on splitting list values and further justify the potential use cases.

Finally there's a [specification](#specification) for the suggestion, which I hope to be more or less feature complete.

Have a look and give some feedback, I for one could use this functionality, and I'm willing to make a PR with whatever we come up with. The only thing I'd like to add would maybe to be able to split on regex as well, but I couldn't come up with a use case where it would make sense.

## Splitting strings

```yaml
context:
  topic: boston/thermometer/723e11c9-75d3-4cef-a921-94304ddf6b1b/temperature
template:
  $let:
    segments: {$eval: split(topic, '/')}
  in:
    city: segments[0]
    type: segments[1]
    uuid: segments[2]
    property: segments[3]
result:
  city: boston
  type: thermometer
  uuid: 723e11c9-75d3-4cef-a921-94304ddf6b1b
  property: temperature
```

```yaml
context:
  time: 9
template:
  $let:
    periods:
      0-5: night
      6-9: morning
      10-12: noon
      13-17: afternoon
      18-23: evening
  in:
    $map: {$eval: periods}
    each(value, key):
      $let:
        range: {$eval: split(key, '-')}
      in:
        $if: "time >= range[0] && time <= range[1]"
        then:
          message: ${time} is in the ${value}
result:
  message: 9 is in the morning
```

## Splitting arrays

```yaml
context:
  Uint8Array: [109, 101, 115, 115, 97, 103, 101, 32, 49, 10, 109, 101, 115, 115, 97, 103, 101, 32, 50]
template:
  $map: {$eval: split(Uint8Array, 10)} # 10 is a newline character used as delimiter in byte stream
  each(section): ${decode(section)}
result:
  - message 1
  - message 2
```

## Crazy

Granted, this example went a little into the future, but it shows how split could be used in a nice
and concise way though 2 additional features are needed in order to achieve this:

- Declarations of tuples from array input would make splitting more sexy _(the `$let` section with `platform` and `version`)_
- An implementation of a string formatter (fx index based on arguments as shown)

```yaml
context:
  dependencies:
    - node:5.2
    - go:1.3
template:
  $let:
    base:
[...]

PR #369, opened 2020-08-20
New feature: `split`

noahlaux:
Introduces `split` to built in string functions enabling string splitting with a delimeter. Input and delimeter can either be `string` or `number`.

Discussed in #368 

# Checklist

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -1940,6 +1940,41 @@ context: {}
 template: {$eval: "lstrip(' \f\n\r\t\vabc \f\n\r\t\v')"}
 result: "abc \f\n\r\t\v"
 ---
+title: split string with delimiter
+context: {}
+template: {$eval: "split('left:right', ':')"}
+result: ['left', 'right']
+---
+title: split string without delimiter
+context: {}
+template: {$eval: "split('spellme', '')"}
+result: ['s', 'p', 'e', 'l', 'l', 'm', 'e']
+---
+title: split with empty string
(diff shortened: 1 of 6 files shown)


specification.yml
```diff
@@ -1940,6 +1940,41 @@ context: {}
...
+---
+title: split string with no match
+context: {}
+template: {$eval: "split('no semicolon', ':')"}
+result: ['no semicolon']
+---
+title: split with number delimiter
+context: {}
+template: {$eval: "split('left1right', 1)"}
+result: ['left', 'right']
```
djmitche, line 1966: Is there a rationale for this particular behavior?
```
>>> 'abc123'.split(1)
  File "<stdin>", line 1, in <module>
TypeError: must be str or None, not int
```

noahlaux: My rationale was that the end user didn't particularly want to think about casting in the context of json-e, and if I don't think there's a huge risk of non deterministic behavior. IMHO, it doesn't matter if the user wanna split on `"1"` or `1`, more the intent.

`split('left1right', 1)` vs `split('left1right', str(1))`

I think the latter just adds noise or I'm not too concerned about failure, though I can be mistaken.

djmitche: What do you think, @helfi92?

helfi92: I agree. We can move forward with the current implementation. Worst case we just revisit this in the future if an issue is raised.

PR #373, opened 2020-08-23
start using GoModules for the Go implementation

djmitche:
This adds `go.mod` and `go.sum`, making this easier to develop without installing it deep in a GOPATH structure.  Since the code doesn't change, versions of Go that are unaware of modules should continue to work just fine.

# Checklist

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -93,9 +93,8 @@ tasks:
             - /bin/bash
             - '-c'
             - >-
-              mkdir -p /go/src/github.com/taskcluster/json-e &&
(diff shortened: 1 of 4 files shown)


petemoore (approved):
It looks sound, but I don't think the CI will pick up any failures, as the new `go.mod` and `go.sum` files are not tested by the CI. Is it worth adding a CI task that builds using go modules, to ensure it works as expected? It could probably happen directly after the non-module build, to reduce setup overhead (since the environment will be almost identical, and the go build step is extremely quick). A separate CI task is also fine with me, I think we only need to test on one platform (e.g. Linux/amd64).

PR #377, opened 2020-08-31
Use valid JSON in tests

djmitche:
This updates the tests to use valid JSON data structures for the tests.  The issue was `{1}`, which at least in Python parses as `{1: None}`, and thus has a non-string as a key.  This isn't allowed in JSON.  [Other parsers](https://onlineyamltools.com/convert-yaml-to-json) seem to translate it `{"1": null}` which is at least valid JSON.  At any rate, we're not building a YAML interpreter here, so let's just use inputs that are valid.

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -4576,22 +4576,22 @@ template: {$eval: 'a != b'}
 result:   false
 ---
 title:    complex object equality, true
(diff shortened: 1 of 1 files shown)


djmitche:
It helps to `git push -f` the final version..

noahlaux (approved):

djmitche:
Woo, thanks!

PR #379, opened 2020-09-24
Test code examples in readme

noahlaux:
This PR is extracting the code block examples in the `readme.md` and integrate them into the javascript test suite, potentially causing it to break in case of errors prior to releasing.

A couple of decision had to be made which can be discussed:

1. **Parsing**
    I had several attempts to extract the tests from the `readme` markdown file.
    I wanted to extract the markdown title closest to the test and put a test title like `$map example 1` etc. Finally, I went with a much simpler solution which parses the markdown with a simple regex to gather the yaml code blocks and makes the stringified template the title, which is a great help when there are errors as you can easily find the right spot.

   **Tries worth mentioning:** 
     - using commonmark to parse the tree and use a walker to generate spec entries
     - using `sed` to generate a static yml file as the current specification file

2. **Using the javascript mocha test only**
    The reasoning behind this is that it's enough to test once as the various language implementation should be similar.

3. **Readme postmortem fixes**
    I fixed the errors that popped up in the readme examples. Some were trivial, some maybe not so. Please have a very deep look.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -825,6 +825,7 @@ template:
   - {$eval: 'lstrip("  room  ")'}
   - {$eval: 'rstrip("  room  ")'}
   - {$eval: 'strip("  room  ")'}
+  - {$eval: 'split("left:right", ":")'}	
 context: {}
 result:
   - "fools!"
@@ -833,6 +834,7 @@ result:
   - "room  "
   - "  room"
   - room
+  - [left, right]
 ```
 
 #### Arrays
--- a/README.md
+++ b/README.md
@@ -436,10 +436,10 @@ result: ["ten"]
 ```
 
 ```yaml
-template: {$match: {"x == 10 || x == 20": "tens", "x == 10": "ten"}}
+# $sort the otherwise arbitrarily ordered results
+template: {$sort:{$match: {"x == 10 || x == 20": "tens", "x == 10": "ten"}}}
 context: {x: 10}
 result: ["ten", "tens"]
-# another possible result: ["tens", "ten"]
 ```
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -825,7 +825,7 @@ template:
   - {$eval: 'lstrip("  room  ")'}
   - {$eval: 'rstrip("  room  ")'}
   - {$eval: 'strip("  room  ")'}
-  - {$eval: 'split("left:right", ":")'}	
+  - {$eval: 'split("left:right", ":")'}
 context: {}
 result:
   - "fools!"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -887,8 +886,8 @@ result:
  - array
  - object
  - function
- - null 
- - ''    # .. which interpolates to an empty string
```
djmitche, line 891: Looks like we changed this in #246 but didn't update docs.  Both of these cases are in `specification.yml` so this fix is correct.

README.md
```diff
@@ -438,9 +438,10 @@ result: ["ten"]
 ```yaml
 template: {$match: {"x == 10 || x == 20": "tens", "x == 10": "ten"}}
 context: {x: 10}
-one possible result: ["tens", "ten"]
-another possible result: ["ten", "tens"]
+result: ["ten", "tens"]
+# another possible result: ["tens", "ten"]
```
djmitche, line 442: This ordering is not guaranteed, at least across languages.  Maybe add a `$sort`?

djmitche (changes requested):
> 1. **Parsing**

Seems fine -- I'm sorry that was a lot of exploration for you!

The titles are surprisingly readable (at least on my wide screen) :)

> 2. **Using the javascript mocha test only**
> The reasoning behind this is that it's enough to test once as the various language implementation should be similar.

Agreed.

> 3. **Readme postmortem fixes**
> I fixed the errors that popped up in the readme examples. Some were trivial, some maybe not so. Please have a very deep look.

Just one issue here, with a pretty easy fix.

README.md
noahlaux: Well, on a positive note, this scenario wouldn't happen again if those tests are run :)

README.md
noahlaux: I understand where you're coming from! Personally, I don't think the `$sort` is needed here in the context of the example, also with the commented out alternative result as an explainer. We could add that explanation to the comment?

djmitche: I am just worried that someday, running in a different version of Node, or in a browser, or whatever, those results will be in the opposite order, and the tests will fail.  I agree that `$sort` is confusing as this is not an example of that operator, but perhaps a comment could help:

```yaml
template:
  # $sort the otherwise arbitrarily ordered results
  $sort: {$match: {"x == 10 || x == 20": "tens", "x == 10": "ten"}} 
```

alternatively, maybe we should define $match to *always* sort its output, thereby avoiding "undefined order" entirely?

noahlaux: I think the latter might be a discussion for another issue, but I like your comprise with the added sorting and comment. I'll add it!

djmitche (approved):
Looks good!

PR #383, opened 2020-10-08
Improve syntax errors for unexpected tokens

djmitche:
..and add a few tests that came up in writing the rust port.

# Checklist


imbstack (commented):
seems like a reasonable idea to me. only thing is with python maybe forgetting to update that string if we add a new one?

djmitche:
The tests would catch that situation.  Golang was left out because it doesn't check error messages.

PR #387, opened 2021-02-11
Add a Rust implementation

### LINKED ISSUE #289: Make a rust implementation

djmitche:
This is largely the work of @owlishDeveloper and @alopezz, as shown in the individual commits.

# Checklist

Fixes #289.


specification.yml
```diff
@@ -32,6 +32,11 @@ context:  {}
 template: {$fromNow: '2 days 3 hours'}
 result:   '2017-01-21T19:27:20.974Z'
 ---
+title:    $eval without context
```
imbstack, line 35: surprised we didn't have this before now!

specification.yml
```diff
@@ -1531,40 +1509,20 @@ template: {$eval: 'floor(key)'}
...
-template: {$eval: 'floor(2.003)'}
-result: 2.0
----
-title: floor (5)
+title: floor (3)
 context: {key: 1}
 template: {$eval: 'floor(key)'}
 result: 1
 ---
-title: floor (6)
```
imbstack, line 1553: why all of these specs removed?

imbstack (commented):
I'm still reviewing most of this but wanted to ask now about the removals of a lot of specifications?

specification.yml
alopezz: I discussed this with Dustin and none of us could figure out what these tests were actually testing. In his words:

> so in combination that's testing that the JSON library considers 3 and 3.0 the same.

djmitche: "combination" there meaning, the deleted test case in combination with the previous test case.  The inputs are the same, so it's just checking the equality operator on JSON objects (which in this case is `serde_json`, and in this case does not consider 3 to equal 3.0, but that's not our problem to solve).

src/value.rs
```diff
@@ -0,0 +1,346 @@
...
+            (Function::WithContext(f1), Function::WithContext(f2)) => {
+                *f1 as *const () == *f2 as *const ()
+            }
+            _ => false,
+        }
+    }
+}
+
+impl Function {
+    // TODO: pub(crate) constructor, take name for Debug
```
imbstack, line 39: should this TODO be done before landing or converted into an issue? linking it here if the latter

imbstack (approved):
Very impressive!

src/value.rs
djmitche: I just fixed it .. I'll merge if it's green!

PR #400, opened 2021-02-22
Remove isJSON util from JavaScript implementation.

mkantor:
The `isJSON` util was calling a nonexistent function (due to a typo).

**UPDATE:** Since it's not actually used, `isJSON` has now been entirely removed.

# Checklist

### DIFF
--- a/src/builtins.js
+++ b/src/builtins.js
@@ -2,7 +2,7 @@ var {BuiltinError} = require('./error');
 var fromNow = require('./from-now');
 var {
   isString, isNumber, isBool,
-  isArray, isObject, isJSON,
+  isArray, isObject,
   isNull, isFunction,
 } = require('./type-utils');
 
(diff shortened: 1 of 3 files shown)


mkantor:
This util is never actually used anywhere and isn't public, so alternatively I could just remove it. Let me know if you'd prefer that.

djmitche:
> This util is never actually used anywhere and isn't public, so alternatively I could just remove it. Let me know if you'd prefer that.

I was just looking into that, and reached the same conclusion.  Yes, let's just remove it.

Thanks for having such a deep look at the source!

mkantor:
Done! f963563 (I'm happy to squash/start a new branch too if you like a clean history).

> Thanks for having such a deep look at the source!

As an experiment I started porting this library to native TypeScript (vs JavaScript with separate type declarations), and it's been finding a few issues like this. The port is nowhere near done yet, but if you have any interest in eventually upstreaming that work let me know and we can talk through the details.

djmitche:
I don't know much about TS, but I'm happy to have it in place and learn :)

djmitche (approved):

PR #404, opened 2021-03-12
Move languages to subdirectories

### LINKED ISSUE #401: Split languages into subdirectories
The JS and Rust implementations both hide their source in `src/`. Python's is under `jsone`. Golang is just in the root.

Let's move that around so that the language name is the top-level directory in the repo.  That will require some sleight-of-hand to keep Go compatibility, but shouldn't be too hard.

djmitche:
This moves all language implementations to subdirectories.  I chose `internal` instead of `go` as it is a special name that will prevent external packages from accessing the "public" symbols within.

# Checklist

Fixes #401.


djmitche:
A change like this, and it passed CI on the first try ?!

imbstack (approved):
thanks, good idea

PR #409, opened 2021-04-28
Give a clear error message when context is not an object

djmitche:
This was never allowed, but the loose type checking in Python and JS
would have allowed some non-objects to be passed.

I believe this is not actually a breaking change, as this never worked in any language.  Before this PR:
```
lamport ~/p/json-e/py [main] $ python
Python 3.6.9 (default, Jan 26 2021, 15:33:00) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import jsone
>>> jsone.render({}, "abcd")
  File "<stdin>", line 1, in <module>
  File "/home/dustin/p/json-e/py/jsone/__init__.py", line 17, in render
    full_context.update(context)
ValueError: dictionary update sequence element #0 has length 1; 2 is required
>>> 
```
(because `update` expects either a dictionary or a sequence of 2-tuples, and instead got a sequence of 1-character strings)
```
lamport ~/p/json-e/js [main] $ node
Welcome to Node.js v15.9.0.
Type ".help" for more information.
> require('./src')({}, "abc");
Uncaught TemplateError: top level keys of context must follow /[a-zA-Z_][a-zA-Z0-9_]*/
    at module.exports (/home/dustin/p/json-e/js/src/index.js:480:11) {
  location: []
}
> 
```
(because `Object.keys("abc")` is `[0, 1, 2]`)

# Checklist

### DIFF
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -475,6 +475,9 @@ let parseUntilTerminator = (source, terminator, context) => {
 };
 
 module.exports = (template, context = {}) => {
+  if (typeof context !== 'object') {
+    throw new TemplateError('context must be an object');
+  }
   let test = Object.keys(context).every(v => /^[a-zA-Z_][a-zA-Z0-9_]*$/.test(v));
   if (!test) {
     throw new TemplateError('top level keys of context must follow /[a-zA-Z_][a-zA-Z0-9_]*/');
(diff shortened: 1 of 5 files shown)


git-developer:
Is it allowed for the context to be an array (which is an object)?

djmitche:
In JSON terms, arrays and objects are entirely different things (unlike in JS, where you're correct that Array inherits Object)

petemoore (approved):
lgtm, thanks dustin!

PR #417, opened 2021-05-17
Forbid uncalled functions in evaluated output

### LINKED ISSUE #158: `$json` should handle functions with an error -`{$json: {$eval: 'min'}}`

mkantor:
Emit an error on uncalled functions in all implementations.

Fixes #158.

# Checklist

### DIFF
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -8,7 +8,7 @@ var stringify = require('json-stable-stringify-without-jsonify');
 var {
   isString, isNumber, isBool,
   isArray, isObject,
-   isTruthy
+  isTruthy, isFunction,
 } = require('./type-utils');
 var addBuiltins = require('./builtins');
 var {JSONTemplateError, TemplateError, SyntaxError} = require('./error');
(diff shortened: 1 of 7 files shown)


mkantor:
Unfortunately because functions can be passed around as regular values during evaluation (such as in [the "dotted function expression" test case](https://github.com/taskcluster/json-e/blob/63c6a7f6d42eced73f5cf0294e873faeb76acb3d/specification.yml#L4255-L4258)) the interpreters can't simply abort when they encounter an uncalled function. The best I was able to come up with was to recursively validate the final evaluated object, but if anyone has other ideas I'm all ears.

js/src/index.js
```diff
@@ -474,6 +478,25 @@ let parseUntilTerminator = (source, terminator, context) => {
...
 };
 
+let containsFunctions = (rendered) => {
+  if (isFunction(rendered)) {
+    return true;
+  } else if (Array.isArray(rendered)) {
+    return rendered.some(containsFunctions);
+  } else if (typeof rendered === 'object' && rendered !== null) {
+    for (const key in rendered) {
+      if (rendered.hasOwnProperty(key)) {
```
djmitche, line 488: another option for this kind of loop is `for (const key of Object.keys(rendered))`

djmitche (approved):
This looks good!  If you would like to make the change below, let me know, but I'm also happy to merge it as-is too.

js/src/index.js
mkantor: Thanks, I made the suggested edit and force-pushed this branch.

djmitche (approved):

djmitche:
Fantastic, thank you for that!

PR #429, opened 2021-11-03
Hide {} in error message from formatting machinery

dtolnay:
I am considering supporting https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html in a future version of the bail macro, which would make e.g. `{v}` interpolate a variable called `v`, matching the standard library macros:

```rust
bail!("error {v}");  // equivalent to bail!("error {}", v)
```

I've been looking into how widely used `{` is in bail macro error messages. It looks like this is one of only very few places where it's used intentionally as a `{` character, as opposed to someone writing `{v}` in the message and expecting it to interpolate the value of `v` (not realizing it doesn't do that currently). If we can write this one a different way, I think it would be reasonable to provide the interpolating behavior from RFC 2795 in the future.

# Checklist

### DIFF
--- a/rs/src/render.rs
+++ b/rs/src/render.rs
@@ -99,7 +99,9 @@ fn interpolate(mut source: &str, context: &Context) -> Result<String> {
                     let expr = source.get(offset + 2..).unwrap();
                     let (parsed, remainder) = interpreter::parse_partial(expr)?;
                     if remainder.get(0..1) != Some("}") {
-                        bail!("unterminated ${..} expression");
+                        // Hide '{' in this error message from the formatting machinery in bail macro
+                        let msg = "unterminated ${..} expression";
+                        bail!(msg);
                     }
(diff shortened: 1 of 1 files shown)


djmitche (approved):
This looks fine.  Would escaping with something like `${{..}}` work, too?

dtolnay:
Yes, but only after `bail!` begins feeding the string to std::fmt. Writing that now would give you double `{{`/`}}` in the message since the newest version doesn't do that yet.

I can follow up with a PR upgrading `anyhow` and changing to `${{..}}` whenever that's ready.

djmitche:
Ah, that makes sense.  I'm happy to just merge this as-is.  I'm excited for this variable-capture thing :)

djmitche:
Oh, right, I'm not a maintainer of this anymore :)

Paging @petemoore!

petemoore (approved):

Issue #430, opened 2021-11-04
Move to a new Github org


djmitche:
This package is used by more than just Taskcluster.

Let's move it to `https://github.com/json-e/json-e` where we can manage ownership independently of Mozilla.  I think that means:

 * Create the new org
 * Add ownership to some Mozilla folks
 * Transfer the repo to the new org

We'll need to do some work to point https://json-e.js.org/ to the new GH pages, but we can sort that out.

jwhitlock:
Thanks for setting up https://github.com/json-e, @djmitche .

One issue @petemoore brought up was that the Go install package is tied to the repository URL. Does anyone have experience moving a Go package?

Looking at https://github.com/taskcluster/json-e/blob/main/RELEASING.md, I believe the others (JS, Python, Rust) can be updated by changing files after moving the repository.

djmitche:
So, there are three possibilities for the Go package that I can think of:
 * Move it, and leave the old version behind, with a doc update
 * Maybe Go will follow the GitHub redirects that are put in place when a repo is moved?
 * We could replace the old repo with "stubs" that just import and re-export names from the latest version of the new repo.

The third option sounds attractive, but it's really no different from the first once we get beyond v4.  I can do a quick experiment to see what happens with the second option.

djmitche:
I can confirm that go does not follow the redirects.

djmitche:
So, here's what I think we should do:
 * Move the repo
 * Immediately fork the new location back to the old location
 * Update https://github.com/taskcluster/json-e/blob/main/jsone.go to import from the new location and update the README to contain only a note about the new location
 * Make a new patch-version tag at the old location
 * Verify that appears on pkg.go.dev
 * Update all references (links, etc.) in the new location
 * Make another patch release from the new location

I'm also a little concerned about the docs -- it looks like "personal" orgs [don't get GitHub pages anymore](https://docs.github.com/en/pages/getting-started-with-github-pages/changing-the-visibility-of-your-github-pages-site) (thanks Microsoft!) so we'll need to find some other way of hosting the docs.  That can probably be an S3 bucket.  I can set up some scripting for that.  I'm not sure that deleting the old repo and re-creating it wouldn't also disable pages, so let's not click any buttons until that scripting is done.

djmitche:
Oh, I appear to have misunderstood that - https://json-e.github.io/_testing/ works fine.  So, I think we'll just need to make a js.org PR.

When should we do all of this?

[...]

PR #432, opened 2021-11-12
update references to new repo

djmitche:
This changes references from the old (taskcluster/json-e) repo to the new (json-e/json-e) repo.  See #430.

I'll keep this PR open to test the TC integration, too.


djmitche:
Let's see if CI runs when I merge this :)

PR #436, opened 2022-01-16
New docs site

djmitche:
I'm working on rewriting the docs site to use mdbook.  I've done a bit of work already to make the playground work and to render examples nicely.

The one thing this doesn't do is allow linking to the playground.  For the moment, I think it's OK to leave that out for now.  We can add such support later.


djmitche:
Hm, the dark/light theming is kind of a mess here.  I'll keep hacking on it.

djmitche:
OK, here's an updated version.  I pretty much left the existing mdbook themes in place, but added a touch of "JSON-e green" here and there.  It's deployed on https://json-e.js.org/ now.

A neat side benefit here, is that the library is now usable from a CDN such as jsdelivr, as it builds js/dist/index.js that rolls up the entire implementation into one file.

TODO.txt
```diff
@@ -0,0 +1,3 @@
+* eliminate deps with https://cdn.jsdelivr.net/npm/json-e@4.4.3 and similar
```
jwhitlock, line 1: this file could be `TODO.md`, looks like Markdown

jwhitlock (approved):
Nice work @djmitche!

TODO.txt
djmitche: Oh, how did I miss that -- that's just my notes.  I'll rebase that out and merge.

PR #439, opened 2022-01-31
upgrade mocha

djmitche:
This should allow updates to ansi-regex and nanoid, both of which have active security alerts out.

These are testing dependencies only (like mocha), so no great harm from the alerts -- but nice to fix it anyway.


djmitche:
Rebased..

jwhitlock (approved):

PR #444, opened 2022-05-07
add wheels to release instructions

### LINKED ISSUE #443: Distribute python wheels on PyPI
> thanks for json-e! :heart: 

## Elevator Pitch
It would be lovely if there were canonical `.whl` packages along with the `.tar.gz` files on PyPI.

## Use Case
My particular interest: wheels (specifically ones ending in `py3-none-any.whl`) are required for [pyodide](https://github.com/pyodide/pyodide)... and could even be used to deliver an in-browser sandbox for pure python stacks like json-e's. 

## Proposal
- where someone/somerobot is running the equivalent of `python setup.py sdist`, add `bdist_wheel`
- where someone/somerobot runs `twine upload dist/*.tar.gz` add `dist/*.whl`

## Background
Practically, recent `pip` versions will just build a wheel and install it, so it's really just pushing more work onto users' systems, and would decrease entropy and increase hashability by just a bit if done here.

More broadly, wheels are the generally-preferred format for python library distribution, as they don't require arbitrary execution at install time.

## Additional Context
Of note: wheel distributions definitely shouldn't _replace_ `sdist` distributions, as they are very handy for downstream re-packagers, as wheels often _don't_ include enough information to re-build a working package... not super relevant in this case, but just to avoid any surprises (disclaimer: maintainer on [conda-forge](https://github.com/conda-forge/json-e-feedstock/) :blush:).

djmitche:
That was pretty easy :)

Fixes #443.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -9,8 +9,9 @@
 * Push to release the JS version -- `git push && git push --tags`
 * Release to PyPi:
   * `cd py/`
-  * `python setup.py sdist`
-  * `twine upload dist/json-e-<version>.tar.gz`
+  * (if you haven't already installed it..) `pip install wheel`
+  * `python setup.py sdist bdist_wheel`
+  * `twine upload dist/json-e-<version>.{tar.gz,whl}`
 * Release to npm:
   * `npm publish` in `js/`
 * Release to crates.io:


alopezz (approved):
👍 Looks like you already released the latest version as a wheel, too, right? Also, any clue why the CI seems to be broken?

djmitche:
I think this was an issue with the git repo URL being used (unauthenticated `git:..`).

PR #448, opened 2022-06-05
Use `html_url` for cloning

djmitche:
See [the webhook docs](https://docs.github.com/en/rest/pulls/pulls) -
this is the only URL that can be cloned from without authentication.

With this change, the `git_url` and `url` vars are the same so I
combined them.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -12,12 +12,10 @@ tasks:
     repo:
       $if: 'tasks_for == "github-push"'
       then:
-        git_url: ${event.repository.url}
         url: ${event.repository.url}
         ref: ${event.after}
       else:
-        git_url: ${event.pull_request.head.repo.git_url}
-        url: ${event.pull_request.head.repo.url}
+        url: ${event.pull_request.head.repo.html_url}
         ref: ${event.pull_request.head.sha}
   in:
     $let:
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -12,12 +12,10 @@ tasks:
     repo:
       $if: 'tasks_for == "github-push"'
       then:
-        git_url: ${event.repository.url}
         url: ${event.repository.url}
         ref: ${event.after}
       else:
-        git_url: ${event.pull_request.head.repo.git_url}
-        url: ${event.pull_request.head.repo.url}
+        url: ${event.pull_request.head.repo.html_url}
```
alopezz, line 18: > With this change, the git_url and url vars are the same so I
combined them.

I don't know the details and probably it doesn't matter much, but just in case I'll point out that the original value for `url` was the `.url` field and not `.html_url`.

alopezz (approved):

.taskcluster.yml
djmitche: I think that was wrong, too.  From [the docs](https://docs.github.com/en/rest/pulls/pulls):
```
        "url": "https://api.github.com/repos/octocat/Hello-World",
```
this is only used in the task description, so it's probably been wrong forever and we never noticed.

Issue #455, opened 2022-10-26
Support a `default` arm in $switch statements


ahal:
This came out of discussion from issue #454 .

To step back a bit, the problem I'd like to solve is the fact that json-e templates tend to get a lot of indentation when you need to check multiple conditions in an $if statement, e.g:
```
$if: val == "a"
then: foo
else:
  $if: val == "b"
  then: bar
  else: baz
```
Every possible value that needs to be checked requires another level of indentation. Dustin pointed out there is a `$switch` operator which would indeed help solve this in some cases. However, because there is no default arm, there is no easy way to implement an `else` clause with it. The above could be re-written as:
```
$let:
  result:
    $switch:
      val == "a": foo
      val == "b": bar
in:
  $if: defined("result")
  then:
    $eval: result
  else: baz
```
But this is even more verbose and still adds an extra level of indentation. Instead, I propose we implement a `$default` arm of switch statements. The previous examples could then be implemented like:
```
$switch:
  val == "a": foo
  val == "b": bar
  $default: baz
```

djmitche:
This sounds like a great proposal :D

noahlaux:
As author of the original `$switch` I felt obligated to implement this good idea that I also have missed myself a couple of times

noahlaux:
I suppose we can close this one when #456 has been merged

PR #456, opened 2022-12-04
`$default` arm  in `$switch` operator

noahlaux:
This change introduces a `$default` fallback case in the `$switch` operator as suggested in #455.

**NB** I'm not fluent in RUST so there might be some street smarts that can be applied! Please suggest, I promise you wont hurt my feelings!

# Checklist

* [X] All tests pass for you on your machine for all implementations (all implementations must behave identically)
* [X] New tests have been added for any new functionality or to prevent regressions of a bugfix
* [X] Added a short description of the change to `newsfragments/$issue.bugfix` (for fixes) or `.feature` (for new features) or `.doc` (for documentation-only changes)

### DIFF
--- a/rs/src/render.rs
+++ b/rs/src/render.rs
@@ -466,7 +466,7 @@ fn match_operator(
 ) -> Result<Value> {
     check_operator_properties(operator, object, |_| false)?;
     if let Value::Object(obj) = _render(value, context)? {
-        get_matching_conditions(obj, context)
+        get_matching_conditions(obj, context, false)
             .or(Err(template_error!("parsing error in $match condition")))
     } else {
         Err(template_error!("$match can evaluate objects only"))
@@ -479,15 +479,9 @@ fn switch_operator(
     object: &Object,
     context: &Context,
 ) -> Result<Value> {
-    check_operator_properties(operator, object, |_| false)?;
-
-    let mut value_without_default = value.clone();
-    if let Value::Object(ref mut o) = value_without_default {
-        o.remove("$default");
-    }
-    
-    if let Value::Object(obj) = _render(&value_without_default, context)? {
-        if let Ok(Value::Array(mut matches)) = get_matching_conditions(obj, context) {
+    check_operator_properties(operator, object, |_| false)?;    
+    if let Value::Object(obj) = _render(value, context)? {
+        if let Ok(Value::Array(mut matches)) = get_matching_conditions(obj, context, true) {
             if let Value::Object(ref o) = value {
                 if let Some(default) = o.get("$default") {
                     let value = _render(default, context)?;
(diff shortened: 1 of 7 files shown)


rs/src/render.rs
```diff
@@ -480,9 +480,21 @@ fn switch_operator(
     context: &Context,
 ) -> Result<Value> {
     check_operator_properties(operator, object, |_| false)?;
-    if let Value::Object(obj) = _render(value, context)? {
+
+    let mut value_without_default = value.clone();
```
djmitche, line 484: Is there a way to avoid this `clone`?  Perhaps by modifying `get_matching_conditions` to skip `$default`?

It looks like there might be an unrelated bug here, which is that the Rust implementation eagerly renders all of the object values, while the other implementations only render the value with the truthy key.  But that could be addressed in another PR.

djmitche (changes requested):
This looks good, with just a performance note for the Rust implementation.

rs/src/render.rs
djmitche: Hm, I didn't realize that either.  It seems it's used for $match and $switch, which are pretty similar.  Maybe add an `ignore_default: bool`?

noahlaux: fixed in https://github.com/json-e/json-e/pull/456/commits/af38ab24452ed2e380c44ea353a64cdf454fda40

djmitche:
Hm, those rust tests are actually failing, but I see that CI is only running `mdbook test`.  There are also three rust tests failing on `main`! :(

I can make up a PR to fix that.  But, looking at the failures, I think to support $default we may need to solve the other issue I noted -- `_render(..)`ing the entire object before checking the conditions.

noahlaux:
ahh, I was relying on the CI tests, I can see now when running the tests locally, sorry :(

djmitche:
No worries!  https://github.com/json-e/json-e/pull/457 should fix CI.

noahlaux:
@djmitche Do you think the eager rendering is part of another PR? It also takes place in the `$match` operator?

djmitche:
If it needs to be solved to fix this issue (and I think that's the case, but maybe not!) then it's OK to do it in this PR.

djmitche:
I'm happy to help, too..

noahlaux:
cool, so basically I think my hurdle would be deferring the rendering without plastering with clones and unwrapping, and also looking into the nice simplicity of the pattern matching in fx the `$if` operator.

djmitche:
I think one important step would be to make `get_matching_conditions` take `&Object` so that the object can still be used later.  It can return `Result<Vec<&Value>>`, I think?  Then, if it returns something of interest, it's up to the caller to call `_render(v)` for each `&Value` in that vector.

djmitche:
@noahlaux I made the rust refactor -- let me know what you think?

noahlaux:
Hey @djmitche, I was out traveling and didn't check here :( sorry for not communicating properly. 

Your refactor is stellar, way better than something I could come up with! Let's go with that and thanks for chipping in!

djmitche:
Awesome, thanks for taking a look!  It's always more fun to collaborate :)

djmitche (approved):

noahlaux:
@djmitche exactly! Merry Christmas to you (if you celebrate)

ahal:
Thank you very much for implementing this! Will result in a huge improvement in our `.taskcluster.yml` files :)

noahlaux:
You’re welcome, I have quite some files that will benefit as well

djmitche:
Now available in json-e 4.5.0.

PR #457, opened 2022-12-06
Actually really run Rust CI

djmitche:
The fix to CI was simple -- it was just a collision of test names, which appeared to GitHub as the same check.

The fix to the failing tests was also simple.


noahlaux (approved):
Great!

noahlaux (approved):

PR #458, opened 2022-12-19
Slightly speedup python tokenizer

serge-sans-paille:
It's always faster to use filtered list comprehension instead of list + filter. As the tokenizer is performance critical and the change doesn't impact readability, it's worth making the move.


djmitche (approved):
Nice fix -- thank you very much!

PR #460, opened 2023-02-04
Fix rust specification.yml rebuild logic

rwedler:
Cargo always rebuilds this crate right now since ```cargo:rerun-if-changed=PATH``` is pointing to a nonexistent file in the crate distribution. Fix is to better detect the location of ```specification.yml```

To reproduce this issue you can grab the crate from crates.io, extract it, and run ```cargo check``` repeatedly.  Another way to reproduce this issue is to temporarily move ```specification.yml``` from the root of this repo to the ```rs``` folder and run ```cargo check``` repeatedly.

See https://github.com/rust-lang/cargo/issues/6003 which proposes making this type of thing a warning in cargo.

# Checklist

### DIFF
--- a/rs/build.rs
+++ b/rs/build.rs
@@ -10,12 +10,18 @@ use yaml_rust::{Yaml, YamlLoader};
 
 fn main() {
     // request to be re-run whenever ../specification.yml changes
-    println!("cargo:rerun-if-changed=../specification.yml");
+    let dev_path = std::path::Path::new("../specification.yml");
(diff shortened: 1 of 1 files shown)


djmitche (approved):
Thanks for this, and for the explanation!

d-sauter:
@djmitche Is there any way we could get v4.5.1 with this fix?

djmitche:
working on it

PR #463, opened 2023-04-07
docs: handle arbitrary newline placement in yaml

### LINKED ISSUE #461: json-e.js.org $let documentation missing $let in templates
When I go to https://json-e.js.org/operators.html#let, the $let operator seems to be missing from all the templates. Looking at the markdown file source that I assume is being hosted, the $let operator has been in all templates for ~two years https://github.com/json-e/json-e/blob/4c9cd1235cbe91ac0d34d351d2a0469c97197b0c/docs/src/operators.md#let. This left me scratching my head. I ended up using https://github.com/json-e/json-e/blob/main/specification.yml#L679 for $let examples.

zedkipp:
Previously some documentation was not being rendered properly on json-e.js.org because of assumptions about where newlines would be in template/context/result examples. This change adds better handling for multi line docs examples.

$let docs before:
Screenshot from 2023-04-06 17-59-03

$let docs after:
Screenshot from 2023-04-06 17-59-17

fixes #461

# Checklist

 - There aren't any tests for this part of the docs. Attached screenshots of the docs with the change.

### DIFF
--- a/docs/jsone.js
+++ b/docs/jsone.js
@@ -134,9 +134,23 @@ const playground_block = (template, context) => {
 const remove_yaml_key = (key, yaml) => {
   const nl = yaml.indexOf('\n');
   if (nl != -1) {
-    const indented = yaml.slice(nl+1).split('\n');
-    const indent = /^ */.exec(indented[0])[0].length;
-    const dedented = indented.map(line => line.slice(indent));
+    const match = RegExp(`(^${key}: *)([\\s\\S]*)`).exec(yaml);
+    const rm_key = match[1];
+    const value = match[2];
+    const lines = value.split('\n');
+
+    let dedented = [];
+    if (lines[0] == "" && lines.length > 0) {
+      // Value started on a new line after the key, so align to the next line.
+      const first_line_indent = /^ */.exec(lines[1])[0].length;
+      for (let i = 1; i < lines.length; i++) {
+        dedented.push(lines[i].slice(first_line_indent));
+      }
+    } else {
+      // Value starts on this line and may continue on the next line
+      dedented = lines
+    }
+
     return dedented.join('\n');
   } else {
     return new RegExp(`^${key}: *(.*)`).exec(yaml)[1];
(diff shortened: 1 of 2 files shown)


docs/jsone.js
```diff
@@ -134,9 +134,30 @@ const playground_block = (template, context) => {
...
+
+    let dedented = [];
+    if (lines[0] == "" && lines.length > 0) {
+      // Value started on a new line after the key, so align to the next line.
+      const first_line_indent = /^ */.exec(lines[1])[0].length;
+      for (let i = 1; i < lines.length; i++) {
+        dedented.push(lines[i].slice(first_line_indent));
+      }
+    } else {
+      // Value starts on this line, so align considering the key on this line.
```
djmitche, line 150: Hm, this works for
```
template: {$let: {ts: 100, foo: 200},
           in: [{$eval: "ts+foo"}, {$eval: "ts-foo"}, {$eval: "ts*foo"}]}
```
but would lose text from something of the form
```
key: [10,
  20]
```
I think that whenever this happens, we're in "JSON" mode, so whitespace is not siginficant. I wonder if the better approach, just in terms of validity, is to not dedent at all in this case?

Zacho2: Yeah, I think you're right. The downside is the docs aren't the most visually appealing. Which of the two approaches below seems best to you?


Given:
```
key: [10,
  20]
```

Approach 1 (remove key, maintain spacing):
```
     [10,
  20]
```

Approach 2 (remove key, and spacing):
```
[10,
  20]
```

djmitche: Let's go with approach 2. I think this is a bad way to write examples, so if it doesn't look great that's OK -- as long as it's not actually wrong :)

Zacho2: Sounds good. I updated the PR accordingly. An example of the $let docs now:
Screenshot from 2023-04-09 17-13-34

djmitche:
Nice work, thanks!!

djmitche (approved):

djmitche:
Docs updated in https://github.com/json-e/json-e/commit/12bdb5a4fc11fcf1dab6b7fd5ba16d8125b96d16

I'll finish releasing 4.5.2 once I have my credentials handy.

PR #464, opened 2023-04-08
Upgrade to nom 7

### LINKED ISSUE #462: Cargo reports future incompatibility with nom dependecy with rustc 1.68
With rustc 1.68.2, building a project with json-e@4.5.1 shows that the `nom@5.1.2` dependency has code that will be an error in future versions.

Looks like nom needs to be updated to at least major version 7 for this to be addressed. From a quick test, it seems like some code changes to json-e are needed for this dependency update.

Snippet of output of `cargo report future-incompatibilities` from my project:
```
 $ cargo report future-incompatibilities
The following warnings were discovered during the build. These warnings are an
indication that the packages contain code that will become an error in a
future release of Rust. These warnings typically cover changes to close
soundness problems, unintended or undocumented behavior, or critical problems
that cannot be fixed in a backwards-compatible fashion, and are not expected
to be in wide use.

Each warning should contain a link for more information on what the warning
means and how to resolve it.


To solve this problem, you can try the following approaches:


- Some affected dependencies have newer versions available.
You may want to consider updating them to a newer version to see if the issue has been fixed.

nom v5.1.2 has the following newer versions available: 6.0.0-alpha1, 6.0.0-alpha2, 6.0.0-alpha3, 6.0.0-beta1, 6.0.0-beta2, 6.0.0-beta3, 6.0.0-beta4, 6.0.0-beta5, 6.0.0, 6.0.1, 6.1.0, 6.1.1, 6.1.2, 6.2.0, 6.2.1, 7.0.0-alpha1, 7.0.0-alpha2, 7.0.0-alpha3, 7.0.0, 7.1.0, 7.1.1, 7.1.2, 7.1.3


- If the issue is not solved by updating the dependencies, a fix has to be
implemented by those dependencies. You can help with that by notifying the
maintainers of this problem (e.g. by creating a bug report) or by proposing a
fix to the maintainers (e.g. by creating a pull request):

  - nom@5.1.2
  - Repository: https://github.com/Geal/nom
  - Detailed warning command: `cargo report future-incompatibilities --id 22 --package nom@5.1.2`

- If waiting for an upstream fix is not an option, you can use the `[patch]`
section in `Cargo.toml` to use your own version of the dependency. For more
information, see:
https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section

The package `nom v5.1.2` currently triggers the following future incompatibility lints:

> warning: trailing semicolon in macro used in expression position
>    --> /Users/zach/.cargo/registry/src/github.com-1ecc6299db9ec823/nom-5.1.2/src/combinator/macros.rs:509:35
>     |
> 509 |     map!(__impl $i, call!($f), $g);
>     |                                   ^
>     |
>    ::: /Users/zach/.cargo/registry/src/github.com-1ecc6299db9ec823/nom-5.1.2/src/number/complete.rs:595:3
>     |
> 595 |   map!(i, le_u128, |x| x as i128)
>     |   ------------------------------- in this macro invocation
>     |
>     = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
>     = note: for more information, see issue #79813 <https://github.com/rust-lang/rust/issues/79813>
>     = note: macro invocations at the end of a block are treated as expressions
>     = note: to ignore the value produced by the macro, add a semicolon after the invocation of `map`
>     = note: `#[allow(semicolon_in_expressions_from_macros)]` on by default
>     = note: this warning originates in the macro `map` (in Nightly builds, run with -Z macro-backtrace for more info)

<repeated warning: trailing semicolon in macro used in expression position>
```

djmitche:
This avoids some future incompatibilities in nom 5.

Note that this contains a workaround for a bug in nom, https://github.com/rust-bakery/nom/issues/1656.

### DIFF
--- a/rs/src/whitespace.rs
+++ b/rs/src/whitespace.rs
@@ -1,0 +1,12 @@
+use nom::{character::complete::multispace0, error::ParseError, sequence::delimited, IResult};
+
+/// chomp trailing whitespace
+// from https://github.com/Geal/nom/blob/master/doc/nom_recipes.md#whitespace
+pub(crate) fn ws<'a, F, O, E: ParseError<&'a str>>(
+    inner: F,
+) -> impl FnMut(&'a str) -> IResult<&'a str, O, E>
+where
+    F: FnMut(&'a str) -> IResult<&'a str, O, E>,
+{
+    delimited(multispace0, inner, multispace0)
+}
--- a/rs/src/whitespace.rs
+++ b/rs/src/whitespace.rs
@@ -1,6 +1,6 @@
 use nom::{character::complete::multispace0, error::ParseError, sequence::delimited, IResult};
 
-/// chomp trailing whitespace
+/// chomp whitespace at either end of a combinator
 // from https://github.com/Geal/nom/blob/master/doc/nom_recipes.md#whitespace
 pub(crate) fn ws<'a, F, O, E: ParseError<&'a str>>(
     inner: F,
--- a/py/jsone/newsfragments/462.bugfix
+++ b/py/jsone/newsfragments/462.bugfix
@@ -1,0 +1,1 @@
+Upgraded Rust parser-builder `nom` to major version 7.
--- a/rs/Cargo.lock
+++ b/rs/Cargo.lock
@@ -9,40 +9,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ee10e43ae4a853c0a3591d4e2ada1719e553be18199d9da9d4a83f5927c2f5c7"
 
 [[package]]
-name = "arrayvec"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
-
-[[package]]
 name = "autocfg"
 version = "1.0.1"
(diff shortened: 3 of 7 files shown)


djmitche:
Oops, that workaround was incorrect! I've added some comments as to why the `.clone()` is required. This was implicit before (nom was doing it for us), so there's no performance loss.

zedkipp (approved):
Nice cleanup on `mod whitespace`

rs/src/whitespace.rs
```diff
@@ -0,0 +1,12 @@
+use nom::{character::complete::multispace0, error::ParseError, sequence::delimited, IResult};
+
+/// chomp trailing whitespace
```
alopezz, line 3: I believe this should probably be:
```suggestion
/// chomp whitespace at either end of a combinator
```
or similar.

alopezz (commented):
The `move || init.clone()` thing is a bit fuzzy for me (haven't touched Rust in quite a while), but otherwise other than the nit about the comment possibly being wrong this looks good.

djmitche:
Thanks! The short summary is, `nom` used to take a cloneable value as the initial value for folding, and clone it when necessary. As of `nom` 7, it instead takes a callable to create the initial value. So that hidden `clone` that `nom` was doing is now visible, making this PR look like it's doing more clones than before, but it's really not.

alopezz (approved):

PR #479, opened 2023-11-04
More general documentation

djmitche:
What is rendering? What's a context? How is all of this typically used?

### DIFF
--- a/docs/src/language.md
+++ b/docs/src/language.md
@@ -8,15 +8,25 @@ irrelevant to its operation.
 
 ## Rendering
 
-A JSON-e template is rendered with a context to produce a result. The
-template, context, and result are all data structures containing strings,
-numbers, true/false, null, objects, and arrays -- the type of data structure
-you get from parsing JSON.
+A JSON-e template is rendered with a context to produce a result. The template,
+context, and result are all data structures containing strings, numbers,
+true/false, null, objects, and arrays -- the type of data structure you get
+from parsing JSON.
 
 The context is an object, giving values for variables at the top level of the
-template. JSON-e defines a few such variables, too, documented in the
-[Built-Ins](./built-ins.md) section. Some operators, such as `$let`, allow the
-template to define additional variables during rendering.
+template.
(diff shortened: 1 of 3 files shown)


docs/src/language.md
```diff
@@ -5,3 +5,18 @@ The following sections describe the JSON-e language.
...
+
+A JSON-e template is rendered with a context to produce a result. The
+template, context, and result are all data structures containing strings,
+numbers, true/false, null, objects, and arrays -- the type of data structure
+you get from parsing JSON.
+
+The context is an object, giving values for variables at the top level of the
+template. JSON-e defines a few such variables, too, documented in the
+[Built-Ins](./built-ins.md) section. Some operators, such as `$let`, allow the
+template to define additional variables during rendering.
```
gregsdennis, line 19: Is it with mentioning that locally-defined variable override those provided with the context?  ("Scope" might be a good word to use.)

djmitche: Good point, thanks!

gregsdennis (approved):

docs/src/language.md
```diff
@@ -5,3 +5,28 @@ The following sections describe the JSON-e language.
...
 irrelevant to its operation.
+
+## Rendering
+
+A JSON-e template is rendered with a context to produce a result. The template,
+context, and result are all data structures containing strings, numbers,
+true/false, null, objects, and arrays -- the type of data structure you get
+from parsing JSON.
+
+The context is an object, giving values for variables at the top level of the
```
gregsdennis, line 16: > The context is an object

The playground doesn't error for a non-object context if the context is never referenced.

djmitche: Ah,
```js
  if (typeof context !== 'object') {
```
is true for arrays (Array is a class, so an instance is an object).

djmitche: Fixed in #481.

gregsdennis: Ah... JavaScript!

PR #481, opened 2023-11-05
Verify that context is an object

djmitche:
..and fix a bug in JS implementation.


djmitche:
Hm, `go` enforces this at the type level, so it can't appear in specification.yml. I'll add per-language unit tests.

gregsdennis (approved):

jwhitlock (approved):
Thanks @djmitche!

PR #483, opened 2023-11-06
Document abbreviations supported by fromNow

### LINKED ISSUE #482: Document full range of specifiers for `fromNow` and `$fromNow`
The current [docs](https://json-e.js.org/operators.html#fromnow) specify:

> The available units are `day`, `hour`, `minute`, and `second`, for all of which a plural is also accepted.

However, there are tests for `min`, `h`, `week`, `month`/`mo`/`m`, and `year`: https://github.com/json-e/json-e/blob/main/specification.yml#L491-L574

Also, these tests seem to show that the space between the value and the specifier is optional.

The documentation should list all of the interval specifiers that are accepted.

Even better, is there a specification for an interval string that is being followed?

djmitche:
Fixes #482.

### DIFF
--- a/docs/src/operators.md
+++ b/docs/src/operators.md
@@ -109,8 +109,7 @@ result:   [1, 2, 3]
 The `$fromNow` operator is a shorthand for the built-in function `fromNow`. It
 creates a JSON (ISO 8601) datestamp for a time relative to the current time
 (see the `now` builtin, below) or, if `from` is given, relative to that time.
-The offset is specified by a sequence of number/unit pairs in a string. For
-example:
+The offset is specified by a sequence of number/unit pairs in a string.
 
 ```yaml,json-e
 template: {$fromNow: '2 days 1 hour'}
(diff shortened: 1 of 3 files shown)


petemoore (approved):
This looks correct, and thanks for adding the tests!

docs/src/operators.md
```diff
@@ -124,8 +123,18 @@ context:  {}
 result:   '2017-01-19T17:27:20.974Z'
 ```
 
-The available units are `day`, `hour`, `minute`, and `second`, for all of which a plural
-is also accepted.
+The available units are `day`, `hour`, `minute`, and `second`, for all of which
```
gregsdennis, line 126: Having a partial list here seems odd.  I'd either full of this list or just say the full list is below.

djmitche: Good point, thanks! I also noticed that `m` was missing from this list (I copied it from the Taskcluster client).

PR #486, opened 2023-11-08
Update Python versions

### LINKED ISSUE #484: Update to newer versions of Node, Python, Go
Let's refresh the versions of Node, Python, and Go we test against.

Also, delete `py/.tox`

djmitche:
Partially addresses #484


djmitche:
This got a little more complicated because `nose` is abandoned, so I had to rewrite to use `pytest`. Also `pep8` is not a thing anymore.

djmitche:
This diff is quite large because I switched to `black` instead of `pep8`. I'm open to other suggestions! At any rate, it should be reviewable commit-by-commit.

jwhitlock (approved):
Thanks @djmitche!

I've been using this to mark a commit as "ran a new formatting tool":

https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

PR #493, opened 2023-11-19
Add a release script

### LINKED ISSUE #469: Automate releases
Releasing this library is difficult, not least because there are so many languages, and I always seem to mess it up somehow.

Let's automate it.

djmitche:
This fixes #469 by at automating releases locally (not in CI, though).


release.sh
```diff
@@ -0,0 +1,100 @@
...
+upload_py() {
+    cd py
+    rm -rf dist/*
+    python setup.py sdist bdist_wheel
+    twine upload dist/*
+    cd ..
+}
+
+push_git() {
+    git push upstream main:main --follow-tags
```
jwhitlock, line 81: ``--follow-tags`` is new to me. Even after reading the docs, I'm not sure how it differs from ``--tags``

jwhitlock (approved):
Looks good, thanks @djmitche!

release.sh
djmitche: I think `--tags` uploaded _all_ tags, which might be too much. `--follow-tags` just uploads tags attached to commits that are also being pushed.

PR #495, opened 2023-11-19
Use latest versions of Node, Rust

### LINKED ISSUE #484: Update to newer versions of Node, Python, Go

djmitche:
And also do a bit of refactoring of `.taskcluster.yml` to make it easier to specify versions of Node and Python.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -20,134 +20,50 @@ tasks:
   in:
     $let:
       environments:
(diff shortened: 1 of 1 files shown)


jwhitlock (approved):
Nice refactor @djmitche!

PR #496, opened 2023-11-21
do a better job pushing tags

djmitche:
The `--follow-tags` option on its own didn't work to upload the v4.6.0 tag, but I ran this next command manually and it did send the tag.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -78,7 +78,8 @@ upload_py() {
 }
 
 push_git() {
-    git push upstream main:main --follow-tags
+    git push upstream main:main
+    git push upstream --follow-gags "v$version"
 }
 
 # Check stuff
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -78,7 +78,8 @@ upload_py() {
 }
 
 push_git() {
-    git push upstream main:main --follow-tags
+    git push upstream main:main
+    git push upstream --follow-gags "v$version"
```
jwhitlock, line 82:
```suggestion
    git push upstream --follow-tags "v$version"
```

jwhitlock (changes requested):
It's nice to have an automated, battle-tested script. Thanks @djmitche!

One change and we're ready to merge v2.

release.sh
djmitche: lol!

jwhitlock (approved):
👍

PR #501, opened 2023-11-24
fix json encoding of extended unicode

### LINKED ISSUE #500: `$json` escaping isn't right in a few tests
Example:

```yaml
context:  {}
```

I understand this is YAML-encoded right now, but `$json` should be outputting valid JSON text, and `\U0001F809` isn't valid JSON text.  It should be encoded using a surrogate pair: `\uD83E\uDC09`.

This would make the test:

```yaml
context:  {}
result:   "[\"\\u0010\",\"Z\",\"\u0103\",\"\\uD83E\\uDC09\"]"
```

The other test this affects is

```yaml
context:  {}
result:   "{\"UaE\":3,\"Z\U00010000P\":1,\"\U0001F809\u732b\":\"hey\",\"\U000E1FFF\u0103\":2}"
```

gregsdennis:
I'm not set up to run the existing implementations.  These tests were found while implementing in .Net, which has a notoriously strict JSON parser.

Resolves #500 by fixing JSON-escaping of unicode chars in three tests.

# Checklist

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -704,17 +704,17 @@ result:   '[2,5,10]'
 title:    representation of various unicode codepoints are consistent
 context:  {}
 template: {$json: ["\x10", "Z", "\u0103","\U0001F809"]}
-result:   "[\"\\u0010\",\"Z\",\"\u0103\",\"\U0001F809\"]"
+result:   "[\"\\u0010\",\"Z\",\"\u0103\",\"\\uD83E\\uDC09\"]"
 ---
 title:    sorting pairs by unicode key strings sorts lexically by codepoint
 context:  {}
 template: {$json: {"\u3347\u0050": 1, "\uE1FF\u0100": 2, "\u0055\u0045": 3}}
-result:   "{\"\u0055\u0045\":3,\"\u3347\u0050\":1,\"\uE1FF\u0100\":2}"
+result:   "{\"\u0055\u0045\":3,\"\u3347\u0050\":1,\"\\uE1FF\u0100\":2}"
 ---
 title:    sorting pairs by unicode key strings sorts lexically by codepoint, even with chars above base plane
 context:  {}
 template: {$json: {"\U0001F809\u732b":'hey', "Z\U00010000P": 1, "\U000E1FFF\u0103": 2, "UaE": 3}}
-result:   "{\"UaE\":3,\"Z\U00010000P\":1,\"\U0001F809\u732b\":\"hey\",\"\U000E1FFF\u0103\":2}"
+result:   "{\"UaE\":3,\"Z\\uD800\\uDC00P\":1,\"\\uD83E\\uDC09\u732b\":\"hey\",\"\\uDB47\\uDFFF\u0103\":2}"
 ---
 title:    $json with undefined properties
 context:  {}


specification.yml
```diff
@@ -704,17 +704,17 @@ result:   '[2,5,10]'
 title:    representation of various unicode codepoints are consistent
 context:  {}
 template: {$json: ["\x10", "Z", "\u0103","\U0001F809"]}
-result:   "[\"\\u0010\",\"Z\",\"\u0103\",\"\U0001F809\"]"
+result:   "[\"\\u0010\",\"Z\",\"\u0103\",\"\\uD83E\\uDC09\"]"
```
djmitche, line 707: I'm not sure why the JSON encoder should be producing escapes here -- its output is unicode, so why not just include the character directly?

gregsdennis: Strictly speaking, JSON is UTF-8.  Anything outside of that needs to be escaped.  I'm using an "unsafe" encoder which will include UTF-16 chars, but UTF-32 still needs to be escaped using surrogate pairs.

djmitche: There is no Unicode "outside of UTF-8". UTF-8 is an encoding of Unicode, and can encode any Unicode character. An encoder is within its rights to escape anything it would like, but all that's required is escaping `"` and `\` (to `\"` an `\\`).

Be careful with terminology: UTF-8, -16, and -32 are encodings, not character sets. I think by "UTF-16 chars" you mean codepoints from 256-65535?

A little [experiment in Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e83418136fbb8d4df8f3b77e9cc1fdfa) shows that it escapes codepoints 0..0x20, with `\b`, `\t`, `\f`, and `\r` in the appropriate places and `\u####` for the remainder. It also escapes `"` and `\` as above. And that's it. With `ensure_ascii=False`, [Python's `json.dumps`](https://docs.python.org/3/library/json.html#json.dumps) does the same. The following JS snippet will do the same in your browser:
```
for (let i = 0; i < 100; i++) {
  let encoded = JSON.stringify(String.fromCharCode(i));
  if (encoded.startsWith("\"\\")) {
    console.log(`${i} ${encoded}`)
  }
}
```

So there's a bit of a consensus among language implementations to only include those escapes, and not anything further. Is there another encoder you could use, or an option similar to Python's `ensure_ascii`?

gregsdennis: Be assured I'm quite familiar with the JSON specification [RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259#section-8), and in particular, section 8.

> JSON text exchanged between systems that are not part of a closed ecosystem MUST be encoded using UTF-8.

As a library to be used by others, this is certainly not part of a closed ecosystem.

To your point, though, section 7 does state that (mostly) any unicode char can be represented directly.

> All Unicode characters may be placed within the quotation marks, except for the characters that MUST be escaped: quotation mark, reverse solidus, and the control characters (U+0000 through U+001F).

As this is the case, the documentation needs to specify what it expects `$json` to do as it seems different implementations handle generating JSON differently in regard to escaping characters.

Thus, my output of `"[\"\\u0010\",\"Z\",\"\u0103\",\"\\uD83E\\uDC09\"]"` should be considered just as valid as including the the character unescaped (as these tests currently expect).

gregsdennis: If it comes down to it, I'll just skip these tests on the grounds that I'm producing equivalent JSON, even though it's not textually identical.

djmitche: Ah, thanks for the pointer to the "MUST be escaped" section - so all of the implementations I've looked at implement the minimal escaping.

You're correct that the escaping your implementation is doing is _valid_ and the two are _equivalent_. It's just not the same as the other implementations. That could lead to some subtle bugs where, for example, someone develops a template on one of the existing implementations and then renders it on the .NET implementation.

That said, there are other differences between the implementations, and it's a pretty niche case where this difference would actually matter: `$json` used with non-BMP characters, and the resulting string used directly rather than parsed back into data structures. Examples of such use would be comparing for equality or hashing.

I think some of the other differences would be with number encodings. As an example, `-0` is encoded as `0` in JS, Rust and Python, but `-0` in Go.

So, let's remove this test case (the non-BMP character, specifically) and document that there may be differences in the JSON encoding depending on the implementation. I'll update #504.

djmitche:
Closing in favor of #504.

PR #502, opened 2023-11-24
add tests found while developing .net implementation

### LINKED ISSUE #476: [ongoing discussion] Tests found while building an implementation
image

Also, the documentation doesn't state this (that I could find).

gregsdennis:
These are tests that I found while developing a .Net implementation for my `json-everything` project.

Ref: https://github.com/gregsdennis/json-everything/pull/566

I'm not set up to run the existing implementations, but I expect they should handle these tests.

Resolves #476

# Checklist

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -1063,12 +1063,12 @@ error: 'TemplateError: $map on objects expects each(x) to evaluate to an object'
 ---
 title:    $map over object, two vars
 context:  {}
-template:	{ $map: {a: 1, b: 2, c: 3}, 'each(v,k)': {'${k}x': {$eval: 'v + 1'}} }
+template: { $map: {a: 1, b: 2, c: 3}, 'each(v,k)': {'${k}x': {$eval: 'v + 1'}} }
 result:   {ax: 2, bx: 3, cx: 4}
 ---
 title:    $map over object, two vars (overwrite)
 context:  {}
-template:	{ $map: {a: 1, b: 2, c: 3}, 'each(v,k)': {'foo': {$eval: 'v + 1'}} }
+template: { $map: {a: 1, b: 2, c: 3}, 'each(v,k)': {'foo': {$eval: 'v + 1'}} }
 result:   {foo: 4}
 ---
 title:    $map over object, one var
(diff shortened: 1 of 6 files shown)


specification.yml
```diff
@@ -1055,6 +1060,21 @@ template:
...
+---
+title:    $map over object, two vars
+context:  {}
+template: { $map: {a: 1, b: 2, c: 3}, 'each(v,k)': {'${k}x': {$eval: 'v + 1'}} }
+result:   {ax: 2, bx: 3, cx: 4}
+---
+title:    $map over object, two vars (overwrite)
+context:  {}
+template: { $map: {a: 1, b: 2, c: 3}, 'each(v,k)': {'foo': {$eval: 'v + 1'}} }
+result:   {foo: 4}
```
djmitche, line 1072: This is currently non-deterministic, and will fail about 2/3 of test runs in languages that randomize object order. JS maintains object insertion order, so it's passing for the JS implementation.

I think this should be treated as an error, rather than randomly selecting one of the keys to win.  I'm curious what your implementation is doing -- selecting the lexically greatest key? Or does it also maintain object insertion order?

djmitche (commented):
I pushed some commits to fix all of the failing tests except the map-overwrite test.

specification.yml
gregsdennis: My implementation passes all of these new tests.  .Net uses insertion order.

gregsdennis:
All of these tests were checked in the playground, so I guess that's just the JS implementation.

specification.yml
djmitche: Got it. Insertion order isn't generally defined for JSON data structures, so we'll need to find some other way to define this. I'd lean toward making it an error, but we could also consider the keys in lexical order. The latter is the fix we did for $map when we realized it was relying on insertion order.

gregsdennis:
I added another test.  This one ensures the result of a function call can be the target of an accessor.  My implementation doesn't support this currently (will fix), but the playground does.

djmitche:
Hm, maybe we can move that new test to a new PR, since the Rust implementation fails, too?

specification.yml
djmitche: We still need to resolve this..

djmitche (changes requested):

djmitche (approved):

PR #503, opened 2023-11-25
document that match sorts keys lexically

### LINKED ISSUE #491: `$match` description does not say anything about sorting
The [docs for `$match`](https://json-e.js.org/operators.html#match) does say anything about sorting, and yet there is this test:

```yaml
title:    $match, 2 matches, ordered by lexcial sorting of property names
context:  {cond: 3}
template: {$match: {'cond == 3': 2, 'cond == 5': 3, 'cond < 5 && cond > 0': 4, 'true': 17}}
result:   [4, 2, 17]
```

Also, what is the point of sorting in this case?  If anything, shouldn't the results be sorted by the order they are in the template (which, depending on the language you're using, may be indeterministic).

djmitche:
Fixes #491. I don't think this warrants a news item.

### DIFF
--- a/docs/src/operators.md
+++ b/docs/src/operators.md
@@ -234,7 +234,7 @@ result: {ax: 2, bx: 3, cx: 4}
 ## `$match`
 
 The `$match` operator is not dissimilar to pattern matching operators.  It gets
-an object, in which every key is a string expression(s) to evaluate to `true`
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

lotas (approved):
Thanks!

petemoore (approved):

PR #504, opened 2023-11-25
Better unicode-non-escaping tests

### LINKED ISSUE #500: `$json` escaping isn't right in a few tests

djmitche:
The previous form of this test was not clear that the intention was to _not_ escape unicode characters.  See #500 and #501 for discussion.

Fixes #500.

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -701,13 +701,29 @@ context:  {a: 1, b: 2}
 template: {$json: [2, 5, {$eval: 'a + b + 7'}]}
 result:   '[2,5,10]'
 ---
-title:    representation of various unicode codepoints are consistent
+title:    ASCII non-printable character is u-escaped
 context:  {}
-template: {$json: ["\x10", "Z", "\u0103","\U0001F809"]}
-result:   "[\"\\u0010\",\"Z\",\"\u0103\",\"\U0001F809\"]"
+template: {$json: ["\x10"]}
+result:   "[\"\\u0010\"]"
+---
+title:    ASCII printable character is not escaped
+context:  {}
+template: {$json: ["Z"]}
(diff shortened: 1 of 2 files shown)


gregsdennis (approved):

specification.yml
```diff
@@ -701,13 +701,24 @@ context:  {a: 1, b: 2}
 template: {$json: [2, 5, {$eval: 'a + b + 7'}]}
 result:   '[2,5,10]'
 ---
-title:    representation of various unicode codepoints are consistent
+title:    ASCII non-printable character is u-escaped
```
petemoore, line 704: Should this be `\u-escaped`?

petemoore (approved):

specification.yml
djmitche: It would appear with two `\` in the YAML, which might be as confusing as this title. WDYT?

petemoore: Yeah escaping sucks! I guess you can use yaml `|-` directive to avoid needing to escape but that would make it inconsistent with others...

PR #505, opened 2023-11-27
add accessor after function call test

gregsdennis:
Per #502, this test was pulled to a new PR.  I've now fixed my implementation so that this passes.

# Checklist

### DIFF
--- a/rs/src/interpreter/parser.rs
+++ b/rs/src/interpreter/parser.rs
@@ -184,22 +184,23 @@ fn unary_expr(input: &str) -> IResult<&str, Node<'_>> {
     ))(input)
 }
 
-/// An index expression (`x[i]`, `x[a..b]` or `x.p`).  These are left-associative at equal
-/// precedence.
-fn index_expr(input: &str) -> IResult<&str, Node<'_>> {
+/// An index expression (`x[i]`, `x[a..b]` or `x.p`) or function call.  These are left-associative
+/// at equal precedence.
+fn index_or_fn_expr(input: &str) -> IResult<&str, Node<'_>> {
     // An index operation without its left-hand side.  The `fold_multi0` closure attaches
     // these to their LHS's and creates Nodes.
-    enum IndexOp<'a> {
+    enum ExprKind<'a> {
         Index(Box<Node<'a>>),
         Slice(Option<Box<Node<'a>>>, Option<Box<Node<'a>>>),
         Dot(&'a str),
+        Func(Vec<Node<'a>>),
     }
 
-    fn index_op<'a>(input: (&'a str, Node<'a>, &'a str)) -> Result<IndexOp<'a>, ()> {
-        Ok(IndexOp::Index(Box::new(input.1)))
+    fn index_expr<'a>(input: (&'a str, Node<'a>, &'a str)) -> Result<ExprKind<'a>, ()> {
+        Ok(ExprKind::Index(Box::new(input.1)))
     }
 
-    fn slice_node<'a>(
+    fn slice_expr<'a>(
         input: (
             &'a str,
             Option<Node<'a>>,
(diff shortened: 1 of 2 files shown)


alopezz (commented):
I'm super rusty both with rust (no pun intended), with nom and with this code, but can't we make this work by adding `function_expr` as a variant of `unary_expr` for example ([here](https://github.com/json-e/json-e/blob/0a74346766dda7e5f55f00d9a59dd95f6ef3b805/rs/src/interpreter/parser.rs#L181-L183))?

djmitche:
Hi @alopezz, thanks for looking!

I think you're suggesting, instead of the diff in this PR, something like
```patch
diff --git rs/src/interpreter/parser.rs rs/src/interpreter/parser.rs
index 3742790..4c72201 100644
--- rs/src/interpreter/parser.rs
+++ rs/src/interpreter/parser.rs
@@ -175,17 +175,17 @@ fn value(input: &str) -> IResult<&str, Node<'_>> {
 
 /// A unary expression
 fn unary_expr(input: &str) -> IResult<&str, Node<'_>> {
     fn node<'a>(input: (&'a str, Node<'a>)) -> Result<Node<'a>, ()> {
         Ok(Node::Un(input.0, Box::new(input.1)))
     }
     alt((
         map_res(ws(tuple((alt((bang_op, tag("-"), tag("+"))), value))), node),
-        value,
+        function_expr,
     ))(input)
 }
 
 /// An index expression (`x[i]`, `x[a..b]` or `x.p`).  These are left-associative at equal
 /// precedence.
 fn index_expr(input: &str) -> IResult<&str, Node<'_>> {
     // An index operation without its left-hand side.  The `fold_multi0` closure attaches
     // these to their LHS's and creates Nodes.
@@ -256,29 +256,29 @@ fn function_expr(input: &str) -> IResult<&str, Node<'_>> {
             ws(tuple((
                 index_expr,
                 tag("("),
                 separated_list0(ws(tag(",")), expression),
                 tag(")"),
             ))),
             node,
         ),
-        index_expr,
+        value,
     ))(input)
 }
 
 /// Exponentiation is right-associative
 fn exp_expr(input: &str) -> IResult<&str, Node<'_>> {
     fn node<'a>(input: (Node<'a>, &'a str, Node<'a>)) -> Result<Node<'a>, ()> {
         Ok(Node::Op(Box::new(input.0), input.1, Box::new(input.2)))
     }
 
     alt((
-        map_res(tuple((function_expr, tag("**"), exp_expr)), node),
-        function_expr,
+        map_res(tuple((index_expr, tag("**"), exp_expr)), node),
+        index_expr,
     ))(input)
 }
 
 /// Define a simple left-associative binary operation which chains to a
 /// higher-precedence operation.
 macro_rules! binop {
     ($name:ident, $higher_prec:ident, $ops:expr) => {
         fn $name(input: &str) -> IResult<&str, Node<'_>> {
```
and that unfortunately loops and blows up the stack - function_expr calls index_expr calls unary_expr calls function_expr.

I think of `(`, `[`, and `.` as infix operators like `+` and `-`, that just happen to have more syntax to their right side. Precedence matters (`1 + f(x)` should not be parsed as `Func(Sum(1, f), [x])`!), but because two of them have a unique last token, there's no need to consider associativity -- `f()[1].x` can only be parsed as `Dot(Index(Func(f, []), 1), "x")` It only matters for `.`, where `a.b.c` needs to be parsed as `Dot(Dot(a, "b"), "c"))`. So it makes sense to handle them all in the same parser-generator function (same precedence) and to use fold_many0 (to be left-associative).

It took me a while, too, to remind myself how `nom` works and how precedence and associativity are handled!

alopezz (approved):
@djmitche Yeah, that's what more or less I what I was thinking of. I'm a bit more familiar with parser generators based on grammars and I struggle a bit with parser combinators. Your explanation is reasonable anyhow.

PR #509, opened 2023-11-28
add tests for zero-argument functions

### LINKED ISSUE #507: Parameterless functions
Do any of the implementations support parameterless functions?

For example, if I wanted to provide a function that generated a random number, it wouldn't need any parameters.  The template would be `{ $eval: 'rand()' }`, and I'd have to supply `rand` in my context.

Would this be accepted, or do the expression parsers expect at least one parameter?

I can't test this online because all of the built-ins have parameters, and I can't define a function in the playground (that I know of).

djmitche:
Adds some tests as suggested in #507.

Fixes #507.

### DIFF
--- a/internal/misc_test.go
+++ b/internal/misc_test.go
@@ -1,0 +1,18 @@
+package jsone
+
+import (
+	"testing"
+
+	"github.com/json-e/json-e/v4/internal/interpreter"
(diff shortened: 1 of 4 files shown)


petemoore (approved):
Thanks Dustin!

PR #510, opened 2023-11-29
Document that literals are allowed in index exprs

### LINKED ISSUE #492: Accessing members of literal values not in docs
Test (in the suite):

```yaml
title:    string interpolation with object evaluation
context:  {key: 'world', num: 1}
template: {message: 'hello ${{a: "world"}.a}', 'k-${{a: 1}.a + {a: 0}.a}': true}
result:   {message: 'hello world', 'k-1': true}
```

An expression like `{a: "world"}.a` is not in the documentation anywhere.

You have [Object Property Access](https://json-e.js.org/expressions.html#object-property-access), but it doesn't mention that this can work off of a value literal.

This test took me by surprise.  As this kind of support wasn't planned for, I now have to rework a good portion of my expression parsing to support it.

djmitche:
Fixes #492.

### DIFF
--- a/docs/src/expressions.md
+++ b/docs/src/expressions.md
@@ -132,6 +132,14 @@ context: {v: {a: 'apple', b: 'bananna', c: 'carrot'}}
 result: 'applebananna'
 ```
 
+Note that the object can be a literal expression:
+
+```yaml,json-e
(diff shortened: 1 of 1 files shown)


petemoore (approved):
Oh that's neat! TIL. :-)

PR #512, opened 2023-11-29
Valid identifier syntax

### LINKED ISSUE #480: Document and test allowed identifiers for variables (context references) and `.` syntax
The [docs](https://json-e.js.org/expressions.html#context-references) don't give any indication as to what's valid for the dot syntax in bar identifiers.

For example, in `x.a`, what is valid for the `a`?

In JSON Path, we've decided to limit valid identifiers to `^[a-zA-Z_][a-zA-Z0-9_]*$`.  That is, start with alpha or underscore, and the rest must be alphanumeric or underscore.

Exploring in the playground reveals this error:

```
Template: {$eval: 'b.a@4'}
Context:  {b: {'a@4': 5}}

SyntaxError: Unexpected input for 'b.a@4' at '@4'
```

What's valid for this syntax is somewhat discoverable, but it should probably be documented.

### From the tests:

```
title: invalid context
context: {'a b c': 1}
template: {}
error: 'TemplateError: top level keys of context must follow /[a-zA-Z_][a-zA-Z0-9_]*/'
panic: true
```

There aren't any tests for nested properties.

gregsdennis:
Resolves #480 

Adds docs and tests for property identifiers, both in the context object and in accessors.

I expect all implementations should pass these.  I verified most of them in the playground.

# Checklist

### DIFF
--- a/docs/src/expressions.md
+++ b/docs/src/expressions.md
@@ -47,6 +47,11 @@ context: {x: 'quick', z: 'sort'}
 result: ['quick', 'sort', 'quicksort']
 ```
 
+Valid identifiers for context references follow the requirements of many
+programming languages: identifiers may only contain letters, underscores, and
+numbers, but may not start with a number. This does imply that the context
+object may only have keys that meet these requirements.
+
 ## Arithmetic Operations
 
 The usual arithmetic operators are all defined, with typical associativity and
@@ -128,8 +133,8 @@ does not have `prop`, while `obj['prop']` will evaluate to `null`.
 
 ```yaml,json-e
 template: {$eval: 'v.a + v["b"]'}
-context: {v: {a: 'apple', b: 'bananna', c: 'carrot'}}
-result: 'applebananna'
+context: {v: {a: 'apple', b: 'banana', c: 'carrot'}}
+result: 'applebanana'
 ```
 
 Note that the object can be a literal expression:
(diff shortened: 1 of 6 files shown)


docs/src/expressions.md
```diff
@@ -47,6 +47,11 @@ context: {x: 'quick', z: 'sort'}
 result: ['quick', 'sort', 'quicksort']
 ```
 
+Valid identifiers for context references follow the requirements of many
+programming languages: identifiers may only contain letters, underscores, and
+numbers, but may not start with a number. This does imply that the context
+object may only have keys that meet these requirements.
```
djmitche, line 53: The context object _could_ have other keys, but those will be inaccessible from JSON-e expressions. Is that what you meant?

gregsdennis: Actually, it looks like the playground errors if there are invalid top-level keys in the context, even if they're not accessed:

image

gregsdennis:
Looks like rust doesn't like starting context access with underscore, and go just gets the error message wrong for context access with hyphens.

docs/src/expressions.md
djmitche: Ah, good catch!

gregsdennis: I can try to reword this if you want.

djmitche: No that's fine. I'll update the other languages

djmitche:
That should fix it..

djmitche (approved):

PR #518, opened 2023-12-17
Add a few corner-case tests

### LINKED ISSUE #389: [tests] more tests for some $operators
These are some missing test cases noted while we were building the rust implementation:

* $let without in
* $let with empty string key 
* $sort of very large integers
* unexpected trailing characters in $eval
* `{$$foo, $$bar}`

djmitche:
This adds some tests that were suggested in #389, fortunately all passing in every implementation.

Fixes #389.


jwhitlock (approved):
Thanks @djmitche! The test specifications make this easy to test and review.

PR #519, opened 2023-12-21
Improve support for Unicode strings

djmitche:
In particular, all strings are indexed and sliced by codepoint, rather than by byte. This is fairly inefficient, because it involves scanning the strings, so the JS and Go implementations take a shortcut for ASCII-only strings, which are probably the common case.

### DIFF
--- a/specification.yml
+++ b/specification.yml
@@ -4106,6 +4106,11 @@ context: {}
 template: {$eval: '"12345"[-2]'}
 result: '4'
 ---
+title: 'string indexing (unicode)'
+context: {key: "\u2622abc\U0001F601"}
+template: {$eval: '[key[0], key[2], key[-1]]'}
+result: ["\u2622", "b", "\U0001F601"]
+---
 title: 'string indexing (noninteger)'
 context: {}
 template: {$eval: '"12345"[2.5]'}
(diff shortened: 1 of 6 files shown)


matt-boris (commented):
Looking good so far. I'll complete my review of this tomorrow.

djmitche:
Sounds good! It turned out to be more complex than I expected..

specification.yml
```diff
@@ -4551,6 +4566,41 @@ title: 'slicing object'
...
+title: 'string slicing - omitted second index'
+context: {key: "hello"}
+template: {$eval: 'key[3:]'}
+result: "lo"
+---
+title: 'string slicing - negative index'
+context: {key: "hello"}
+template: {$eval: 'key[-3:]'}
+result: "llo"
+---
```
matt-boris, line 4594: Didn't look like you had a test like this.

```suggestion
title: 'string slicing - negative index, omitted second index'
context: {key: "hello"}
template: {$eval: 'key[-3:]'}
result: "llo"
---
title: 'string slicing - negative index'
context: {key: "hello"}
template: {$eval: 'key[-3:4]'}
result: "ll"
---
```

matt-boris (commented):
JS and Go changes look good to me. I'm not proficient in Rust, but can say it _looks_ good.

Added one suggestion for another test 👍🏻

djmitche:
Thanks! It was an interesting thing to fix, with different implementations in different languages that handle unicode differently!

gregsdennis:
JsonE.Net has now been updated to support these new tests.

PR #522, opened 2024-01-27
Enforce that `split` always has two arguments.

### LINKED ISSUE #521: split() docs do not mention that one-argument form is allowed
Identified in #520.

djmitche:
The one-argument form existed for JS, Python, and Go -- but not Rust. But it produced different results for JS (single-element list) and Python/Go (list of single characters).

Fixes #521.


petemoore (approved):
Looks good to me, thanks Dustin!

PR #530, opened 2024-06-25
Removing print statement

da0x:
This commit addresses an issue where an arbitrary line was being injected into the standard output stream. This causes problems for command line tools that rely on pipes to communicate with each other. By removing this line, we ensure proper communication and maintain the integrity of our pipeline.


djmitche:
Curious why you closed this?

PR #531, opened 2024-09-13
Add a `$reduce` operator

tadpol:
This adds a `$reduce` operator.  It is somewhat similar to the `$merge` operator expect that it can produce any valid JSON type, not just objects. (Maybe more like nesting `$merge` and `$map`)

As a simple example:
```yaml
$reduce: [2, 4, 6]
each(acc, x): "${acc}${x}"
initial: 'pre '
```

Will output: `pre 246`

I am submitting this as a draft for now because:
1. I would like some feedback on this operator.
2. ~Only some versions are implemented.~
3. ~No docs yet.~

# Checklist

* [X] All tests pass for you on your machine for all implementations (all implementations must behave identically)
* [X] New tests have been added for any new functionality or to prevent regressions of a bugfix

### DIFF
--- a/docs/src/operators.md
+++ b/docs/src/operators.md
@@ -240,7 +240,7 @@ result: {ax: 2, bx: 3, cx: 4}
 ## `$reduce`
 
 The `$reduce` operator evaluates an expression with each value of the given array and
-the result of the prior expression, reducing the array or object into a single JSON value.
+the result of the prior expression, reducing the array into a single JSON value.
 
 This operation is sometimes called `fold`, `accumulate`, `aggregate`, or `inject`.
 
--- a/docs/src/operators.md
+++ b/docs/src/operators.md
@@ -239,10 +239,10 @@ result: {ax: 2, bx: 3, cx: 4}
 
 ## `$reduce`
 
-The `$reduce` operator evaluates an expression with each value of the given array or object and
+The `$reduce` operator evaluates an expression with each value of the given array and
 the result of the prior expression, reducing the array or object into a single JSON value.
 
-This operation is sometimes called `fold`, `accumulate`, `aggregate`, `reduce`, or `inject`.
+This operation is sometimes called `fold`, `accumulate`, `aggregate`, or `inject`.
 
 An initial result is passed as the accumulator for the first evaluation of the expression.
 
--- a/docs/src/operators.md
+++ b/docs/src/operators.md
@@ -131,6 +131,7 @@ template: {$fromNow: '1 hour', from: '2017-01-19T16:27:20.974Z'}
 context:  {}
 result:   '2017-01-19T17:27:20.974Z'
 ```
+
 The available units, including useful shorthands, are:
 
 ```none
@@ -234,6 +235,38 @@ template:
   each(y): {'${y.key}x': {$eval: 'y.val + 1'}}
 context:  {}
 result: {ax: 2, bx: 3, cx: 4}
+```
+
+## `$reduce`
+
+The `$reduce` operator evaluates an expression with each value of the given array or object and
+the result of the prior expression, reducing the array or object into a single JSON value.
+
+This operation is sometimes called `fold`, `accumulate`, `aggregate`, `reduce`, or `inject`.
+
+An initial result is passed as the accumulator for the first evaluation of the expression.
+
+The `each` function defines the accumulated, or prior result, and the current value.  The result of
+this function will be passed as the accumulated to the next call of `each`.
+
+```yaml,json-e
+template:
+  $reduce: [{name: Apple, price: 1}, {name: Orange, price: 0.75}, {name: Pear, price: 1.1}]
+  initial: 0
+  each(acc, v): {$eval: 'acc + v.price'}
+context:  {}
+result:   2.85
+```
+
+The `each` function can define three variables, in which case the third is the 0-based index of the element.
+
+```yaml,json-e
+template:
+  $reduce: [2, 5, 8]
+  initial: 0
+  each(acc, v, i): {$eval: 'acc + v * 10 ** i'}
+context:  {}
+result:   852
 ```
 
 ## `$find`
@@ -420,6 +453,7 @@ result:   [{a: 1, b: []}, {a: 2}, {a: 3}]
 ```
 
 The sort is stable:
+
 ```yaml,json-e
 template:
   $sort: ["aa", "dd", "ac", "ba", "ab"]
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -219,6 +219,37 @@ operators.$map = (template, context) => {
   }
 };
 
+operators.$reduce = (template, context) => {
+  const EACH_RE = 'each\\(([a-zA-Z_][a-zA-Z0-9_]*),\\s*([a-zA-Z_][a-zA-Z0-9_]*)(,\\s*([a-zA-Z_][a-zA-Z0-9_]*))?\\)';
+  checkUndefinedProperties(template, ['\\$reduce', 'initial', EACH_RE]);
+  let value = render(template['$reduce'], context);
+  if (!isArray(value)) {
+    throw new TemplateError('$reduce value must evaluate to an array');
+  }
+
+  if (Object.keys(template).length !== 3) {
+    throw new TemplateError('$reduce must have exactly three properties');
+  }
+
+  let eachKey = Object.keys(template).find(k => k !== '$reduce' && k !== 'initial');
+  let match = /^each\(([a-zA-Z_][a-zA-Z0-9_]*),\s*([a-zA-Z_][a-zA-Z0-9_]*)(,\s*([a-zA-Z_][a-zA-Z0-9_]*))?\)$/.exec(eachKey);
+  if (!match) {
+    throw new TemplateError('$reduce requires each(identifier) syntax');
+  }
+
+  let a = match[1];
+  let x = match[2];
+  let i = match[4];
+  let each = template[eachKey];
+  let initialValue = template['initial'];
+
+  return value.reduce((acc, v, idx)=>{
+    const args = typeof i !== 'undefined' ? {[a]: acc, [x]: v, [i]: idx} : {[a]: acc, [x]: v};
+    const r = render(each, Object.assign({}, context, args));
+    return r === deleteMarker ? acc : r;
+  }, initialValue);
+};
+
 operators.$find = (template, context) => {
   const EACH_RE = 'each\\(([a-zA-Z_][a-zA-Z0-9_]*)(,\\s*([a-zA-Z_][a-zA-Z0-9_]*))?\\)';
   checkUndefinedProperties(template, ['\\$find', EACH_RE]);
--- a/py/jsone/newsfragments/+reduce.feature
+++ b/py/jsone/newsfragments/+reduce.feature
@@ -1,0 +1,1 @@
+Introduces the `$reduce` operator which combines an array or object into a single value.
--- a/rs/src/op_props.rs
+++ b/rs/src/op_props.rs
@@ -2,7 +2,7 @@ use anyhow::Result;
 use nom::{
     branch::alt,
     bytes::complete::tag,
-    character::complete::{alpha1, alphanumeric1},
+    character::complete::{alpha1, alphanumeric1, multispace0},
     combinator::{map_res, opt, recognize},
     multi::many0,
     sequence::{pair, tuple},
@@ -10,6 +10,7 @@ use nom::{
 };
 
 fn ident(input: &str) -> IResult<&str, &str> {
+    let (input, _) = multispace0(input)?;
     recognize(pair(
         alt((alpha1, tag("_"))),
         many0(alt((alphanumeric1, tag("_")))),
@@ -31,6 +32,26 @@ fn each(input: &str) -> IResult<&str, (&str, Option<&str>)> {
 /// Parse the each(..) property of $map, or return None if no match
 pub(crate) fn parse_each(input: &str) -> Option<(&str, Option<&str>)> {
     match each(input) {
+        Ok(("", r)) => Some(r),
+        _ => None,
+    }
+}
+
+fn each_three(input: &str) -> IResult<&str, (&str, &str, Option<&str>)> {
+    fn to_result<'a>(
+        input: (&str, &'a str, &str, &'a str, Option<(&str, &'a str)>, &str),
+    ) -> Result<(&'a str, &'a str, Option<&'a str>), ()> {
+        Ok((input.1, input.3, input.4.map(|x| x.1)))
+    }
+    map_res(
+        tuple((tag("each("), ident, tag(","), ident, opt(tuple((tag(","), ident))), tag(")"))),
+        to_result,
+    )(input)
+}
+
+/// Parse the each(..) property of $reduce, or return None if no match
+pub(crate) fn parse_each_three(input: &str) -> Option<(&str, &str, Option<&str>)> {
+    match each_three(input) {
         Ok(("", r)) => Some(r),
         _ => None,
     }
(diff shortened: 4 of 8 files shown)


noahlaux:
Hi @tadpol, first of all thanks for contributing to the project! I feel strongly for functional programming so I'm totally with you, I even contributed a couple of operators to json-e which is normally to be found in functional programming languages, fx [find](https://github.com/json-e/json-e/issues/525).

As you can read in the issue, it was possible to actually do what I wanted with the existing operators, but it was slightly cumbersome and verbose. I actually thought about suggesting a `reduce` as well, but I couldn't come up with a good use case for it that wasn't possible just using the existing operators.

I was also thinking about making a filter function, but in fact since json-e omit null values in the [$map](https://json-e.js.org/operators.html#map) operator, the overhead of just adding a [$if](https://json-e.js.org/operators.html#if---then---else) doesn't seem to much of a deal.

```yaml
# A potential filter function
$filter: [1, 2, 3]
each(value): value > 1
```

```yaml
# How you can already do in json-3
$map: [1, 2, 3]
each(value):
  $if: value > 1
  then:
    $eval: value
```

Looking at your (granted) quite simple example, it seems to me like:

1. There are already ways to achieve that with the existing operators and it might even be simpler

  Using existing builtin functionality in json-e:

  ```yaml
  # Example 1 with inline
  pre ${join([2, 4, 6],'')} # => pre 246
  ```

  ```yaml
  # Example 2 with context
  template: pre ${join(value, '')}
  context:
    value: [2, 4, 6]
  ```

  ```yaml
  # Example 3 with $let
  $let:
    value: [2, 3, 4]
  in: pre ${join(value,'')}
  ```

1. I my mind, as a programmer, and knowing reduce from other language, your example should yield `pre 6`. It's unclear why `${x}` is being joined while `${acc}` isn't. I would expect the `each` function to be called with the accumulator and the current value, and the result of the function to be the new accumulator. This is how reduce works in other languages I know of.

Maybe I'm wrong but I'd love to understand how you think the inner workings of the reduce should work and maybe you could come up with some more advanced examples where the existing operators can't do the job.

tadpol:
Agreed, most of the simple examples for `$reduce` can be already done in json-e.  It is just that the simple examples are the easiest to explain how it works.

> I would expect the each function to be called with the accumulator and the current value, and the result of the function to be the new accumulator. This is how reduce works in other languages I know of.

That is how I understand reduce to work and was my intention in the simple example.  Maybe I made I mistake typing it?

Why I'm proposing this is for a more complex use case.  I'll just show the real world I'm dealing with.  I tried for a while to figure out to do this already in json-e. However because of my specific use case, while I could solve it with many nested `$let`s it is visually unappealing.  In most other tooling, a reduce function of some kind is what I would have used. (or worse global variables, or slightly better a captured scope)

The shape of the context in this case is fixed.  So many of the things I'm trying to fix could be solved if I could change the context, but I can't.  There is other stuff in there (and in the template), but for what I want a `$reduce` for, here is just what is relevant.  There are 9 keys, five can hold numbers or undefined, four hold strings.  The four strings are the basic math operations.  (The system that calls the json-e deals with making sure all of the undefined number keys are "at the end", and are aligned with the op keys.)

An example context with all nine keys filled:
```json
{
 "A": 10,
 "B": 2,
 "C": 8,
 "D": 1,
 "E": 1,
 "opA": "+",
 "opB": "+",
 "opC": "/",
 "opD": "+"
}
```

Now I need a bit of json-e to take that mess and produce a single number using the math operations provided.  If the `$reduce` operation existed, I could do this:

```yaml
$let:
  values:
    $map:
      $eval: '[B,C,D,E]'
    'each(v)':
      $if: defined(v)
      then:
        eval: v
  ops:
    $eval: '[opA,opB,opC,opD]'
in:
  $reduce: values
  initial:
    $eval: 'A'
  'each(a,v,i)':
    $switch:
      'ops[i] == "+"':
        $eval: 'a + v'
      'ops[i] == "-"':
        $eval: 'a - v'
      'ops[i] == "*"':
        $eval: 'a * v'
      'ops[i] == "/"':
        $eval: 'a / v'
      $default:
        $eval: a

```

This is still a bit of a mess, but it is cleaner than the four layer nested `$let` mess I have currently.

noahlaux:
@tadpol after reading my answer again yesterday, I had to r remove quite a bit because I had misunderstood the inner workings of your proposal. I'm sorry you had to take time to explain something that was already clear!

Now, with your second example I can see why `reduce` makes sense indeed. I also tried to think about other ways to solve your particular problem, but wasn't able to come up with something better.

djmitche:
I'm happy to see this conversation! As a meta-point, please be careful editing posts, and try to limit those edits to not modify the meaning, and use ~~strikeout~~ to indicate that some parts have been repudiated, e.g., "~~the sky is green~~ [edit: see below where I was reminded that it is, in fact, blue]". Otherwise it's confusing for later readers to see a reply to a message different from the one appearing in the issue.

noahlaux:
> I'm happy to see this conversation! As a meta-point, please be careful editing posts, and try to limit those edits to not modify the meaning, and use ~strikeout~ to indicate that some parts have been repudiated, e.g., "~the sky is green~ [edit: see below where I was reminded that it is, in fact, blue]". Otherwise it's confusing for later readers to see a reply to a message different from the one appearing in the issue.

@djmitche I totally agree! I did realize my wrongful statements rather quickly and I thought I could spare @tadpol to even think about it by removing, but he was very quick :)

js/src/index.js
```diff
@@ -219,6 +219,48 @@ operators.$map = (template, context) => {
...
+  let value = render(template['$reduce'], context);
+  if (!isArray(value) && !isObject(value)) {
+    throw new TemplateError('$reduce value must evaluate to an array or object');
+  }
+
+  if (Object.keys(template).length !== 3) {
+    throw new TemplateError('$reduce must have exactly three properties');
+  }
+
+  let eachKey = Object.keys(template).filter(k => k !== '$reduce' && k !== 'initial')[0];
```
noahlaux, line 234: I suppose you could just use a find here

```suggestion
  let eachKey = Object.keys(template).find(k => k !== '$reduce' && k !== 'initial');
```

js/src/index.js
```diff
@@ -219,6 +219,48 @@ operators.$map = (template, context) => {
...
+      const r = render(each, Object.assign({}, context, args));
+      return r === deleteMarker ? acc : r;
+    }, initialValue)
+  } else {
+    return value.reduce((acc, v, idx)=>{
+      let args = typeof i !== 'undefined' ? {[a]: acc, [x]: v, [i]: idx} : {[a]: acc, [x]: v};
+      const r = render(each, Object.assign({}, context, args));
+      return r === deleteMarker ? acc : r;
+    }, initialValue);
+  }
```
noahlaux, line 261: Keeping it DRY

```suggestion
  value = isObject(value) ? Object.keys(value).map(key => ({key, val: value[key]})) : value;

  return value.reduce((acc, v) => {
    let args = typeof i !== 'undefined' ? {[a]: acc, [x]: v.val, [i]: v.key} : {[a]: acc, [x]: v};
    const r = render(each, Object.assign({}, context, args));
    return r === deleteMarker ? acc : r;
  }, initialValue);
```

noahlaux (changes requested):

py/jsone/newsfragments/+reduce.feature
```diff
@@ -0,0 +1 @@
+Introduces the `$reduce` operator which combines an array or object into a single JSON value.
```
noahlaux, line 1: Introduces the `$reduce` operator which accumulates an array or object into a single value.

docs/src/operators.md
```diff
@@ -236,6 +237,77 @@ context:  {}
...
+
+### Over Objects
+
+The `each` function defines variables for the accumulated, value, and key, in that order.
+
+```yaml,json-e
+template:
+  $reduce: A
+  initial: 10
+  each(acc, v, k):
```
gregsdennis, line 280: There may be some implementation consistency issues here.  Some languages don't iterate over objects in a deterministic way.  We've seen in JSON Schema where iteration over objects produces the children in varying orders even within the same language sometimes.  Depending on the function, that could produce different results with each run.

tadpol: Yes.  I am assuming that when iterating over objects the order is non-deterministic.  And that when using `$reduce` people will take this into account.

noahlaux: Yeah the order can't be trusted, it's also left out in iteration over objects in [maps](https://json-e.js.org/operators.html#over-objects). So maybe for consistency we should leave it out?

docs/src/operators.md
```diff
@@ -236,6 +237,77 @@ context:  {}
 result: {ax: 2, bx: 3, cx: 4}
 ```
 
+## `$reduce`
+
+The `$reduce` operator evaluates an expression with each value of the given array or object and
+the result of the prior expression.  Reducing the array or object into a single JSON value.
```
djmitche, line 243:
```suggestion
the result of the prior expression, reducing the array or object into a single JSON value.
```

docs/src/operators.md
```diff
@@ -236,6 +237,77 @@ context:  {}
...
+
+An initial result is passed in for the first evaluation of the expression.
```
djmitche, line 245:
```suggestion
An initial value is passed as the accumulator for the first evaluation of the expression.
```

docs/src/operators.md
```diff
@@ -236,6 +237,77 @@ context:  {}
...
+result:   852
+```
+
+### Over Objects
+
+The `each` function defines variables for the accumulated, value, and key, in that order.
+
+```yaml,json-e
+template:
+  $reduce: A
```
djmitche, line 278:
```suggestion
  $reduce: {$eval: "A"}
```

docs/src/operators.md
djmitche: It isn't that $map doesn't specify order. Order of object keys is simply not defined, in the sense that two objects are equal if every key/value pair in one is also in the other. So, $map is deterministic: given an object, it produces an object.

We should avoid anywhere where the implementation detail of the order in which iteration occurs over an object has an effect on the results, rather than relying on the template author to ensure this. Otherwise two renderings of the same template with the same context could sometimes have different results, which would cause all manner of intermittent bugs.

In this case, I think that restriction requires either defining the order in which that iteration occurs (sorted keys), or simply not defining the operator over objects. Since we already have $sort, I would lean toward the latter.

docs/src/operators.md
djmitche: Also it might be worth mentioning that this is also known as "fold" or "accumulate", just in case that matches the term from users' preferred language.

rs/src/op_props.rs
```diff
@@ -2,14 +2,15 @@ use anyhow::Result;
...
-    character::complete::{alpha1, alphanumeric1},
+    character::complete::{alpha1, alphanumeric1, multispace0},
     combinator::{map_res, opt, recognize},
     multi::many0,
     sequence::{pair, tuple},
     IResult,
 };
 
 fn ident(input: &str) -> IResult<&str, &str> {
+    let (input, _) = multispace0(input)?;
```
djmitche, line 13: I'm curious why this was required? Were there expressions that didn't parse without it?

specification.yml
```diff
@@ -1102,6 +1102,92 @@ template: {$map: {a: 1, b: 2, c: 3}, 'each(y)': {'${y.key}x': {$eval: 'y.val + 1
...
+---
+title:    reduce skips delete-marker from $if
+context:  {}
+template:
+  $reduce: {a: 2, b: 4, c: 6}
+  initial: 0
+  each(acc, v, k):
+    $if: 'v != 4'
+    then: {$eval: 'acc + v'}
+result: 8
```
djmitche, line 1188: Please add some error cases as well: passing disallowed types to `$reduce`, omitting `initial`, and malformed `each`.

djmitche (changes requested):
This is looking great!

docs/src/operators.md
gregsdennis: "aggregate" is another one I see a lot

docs/src/operators.md
gregsdennis: My concern was heightened when I saw [this comment and use case](https://github.com/json-e/json-e/pull/531#issuecomment-2351144096).  Looking at it again, it looks like all of the orderings are explicitly specified by the embedded arrays, but it's not a long leap to end up with something that depends on key selection order.

djmitche: I believe Python at least randomizes the order of its maps, so if we do introduce a dependency on order hopefully tests will find it.

rs/src/op_props.rs
tadpol: The implementations for js, py, and go all handle the each function for map with spaces after the comma.  Rust is the only one that didn't. So js, py, go: `each(x, i)` or `each(x,i)`, and rust only `each(x,i)`.

Arguably, I can see this as a separate bug that could be in its own PR.

tadpol: …and I'm the type of person that habitually puts a space after a comma, and that's the way I wrote the `$reduce` tests in specification.yaml

docs/src/operators.md
tadpol: It certainly would be clearer to have `$reduce` only work over arrays.  The only reason I included working over objects is that I started the `$reduce` code by copying the `$map` code.

rs/src/op_props.rs
djmitche: Oh, good catch! And, those `$reduce` tests will serve as a regression test for this.

Another issue/PR would have been nice for archaeological purposes (such as someone trying to figure out when this was fixed), but it's done and now it's been discussed here so I think it's fine as-is.

docs/src/operators.md
djmitche: Let's do that (only work over arrays).

If it becomes necessary, we can discuss how to best support `$reduce` over objects at that time.

gregsdennis: @tadpol you marked this as resolved, but I still see iterating over objects in the docs.

tadpol: Oh for. I completely forgot to update the docs. Thanks for catching that.

gregsdennis:
(Force pushes make it difficult to see progression.  But this isn't my repo.)

docs/src/operators.md
```diff
@@ -236,6 +237,38 @@ context:  {}
 result: {ax: 2, bx: 3, cx: 4}
 ```
 
+## `$reduce`
+
+The `$reduce` operator evaluates an expression with each value of the given array or object and
+the result of the prior expression, reducing the array or object into a single JSON value.
```
gregsdennis, line 243: Still has language about iterating over objects.

docs/src/operators.md
```diff
@@ -236,6 +237,38 @@ context:  {}
...
+
+This operation is sometimes called `fold`, `accumulate`, `aggregate`, `reduce`, or `inject`.
```
gregsdennis, line 245: Do we need to repeat "reduce" in this list?

docs/src/operators.md
```diff
@@ -236,6 +237,38 @@ context:  {}
 result: {ax: 2, bx: 3, cx: 4}
 ```
 
+## `$reduce`
+
+The `$reduce` operator evaluates an expression with each value of the given array and
+the result of the prior expression, reducing the array or object into a single JSON value.
```
gregsdennis, line 243: "array or object"

djmitche:
I'm happy to merge once the last vestiges of objects in the doc are eliminated..

gregsdennis (approved):

gregsdennis:
🎉❤️

djmitche (approved):

djmitche:
v4.8.0 is released!

PR #534, opened 2024-09-23
Introducing `range` builtin

kwhitehouse:
# Summary
Implements a new builtin `range` function. See discussion in https://github.com/json-e/json-e/issues/532 for background.

# Details
* Implements support for a new builtin `range` function in javascript, python, rust, golang
* Adds tests in `specification.yml`
* Updates documentation to refer to functionality for the new function

### DIFF
--- a/docs/src/built-ins.md
+++ b/docs/src/built-ins.md
@@ -148,13 +148,29 @@ The `range()` built-in generates an array based on the following inputs:
   this negative value (exclusive).
 * `step` - Optional. An integer specifying a step to apply to each value within
   the range. If not specified, defaults to `1`. Can be negative, but cannot be
-  zero.
-    * For a positive step, the contents of a range r are determined by the
-      formula `r[i] = start + step*i` where `i >= 0` and `r[i] < end`. The
-      resulting range will be empty if `start` >= `end`.
-    * For a negative step, the contents of the range are still determined by the
-      formula `r[i] = start + step*i`, where `i >= 0` and `r[i] > end`. The
-      resulting range will be empty if `start` <= `end`.
+  zero. The contents of a range r are determined by the following formula:
+  ```
+  IF step > 0 THEN
+
+    i = start
+    WHILE i < end:
+      r(i) = i
+      i = i + step
+    END WHILE
+
+  ELSE if step < 0 THEN
+
+    i = start
+    WHILE i > end:
+      r(i) = i
+      i = i + step
+    END WHILE
+
+  END IF
+  ```
+
+Notably, the resulting range will be empty if `start >= end` and `step` is
+positive or if `start` <= `end` and `step` is negative.
 
 ```yaml,json-e
 template:
--- a/docs/src/built-ins.md
+++ b/docs/src/built-ins.md
@@ -153,7 +153,7 @@ r[i] = start + step*i where i >= 0 and r[i] < end.
 For a negative step, the contents of the range are still determined by the
 formula r[i] = start + step*i, where i >= 0 and r[i] > end.
 
-```
+```yaml,json-e
 template:
   $map: {$eval: 'range(1, 5)'}
   each(x): {$eval: 'x'}
--- a/docs/src/built-ins.md
+++ b/docs/src/built-ins.md
@@ -148,29 +148,31 @@ The `range()` built-in generates an array based on the following inputs:
   this negative value (exclusive).
 * `step` - Optional. An integer specifying a step to apply to each value within
   the range. If not specified, defaults to `1`. Can be negative, but cannot be
-  zero. The contents of a range r are determined by the following formula:
-  ```
-  IF step > 0 THEN
+  zero.
 
-    i = start
-    WHILE i < end:
-      r(i) = i
-      i = i + step
-    END WHILE
+The contents of a range r are determined by the following formula:
+```
+IF step > 0 THEN
 
-  ELSE if step < 0 THEN
+  i = start
+  WHILE i < end:
+    r(i) = i
+    i = i + step
+  END WHILE
 
-    i = start
-    WHILE i > end:
-      r(i) = i
-      i = i + step
-    END WHILE
+ELSE if step < 0 THEN
 
-  END IF
-  ```
+  i = start
+  WHILE i > end:
+    r(i) = i
+    i = i + step
+  END WHILE
+
+END IF
+```
 
 Notably, the resulting range will be empty if `start >= end` and `step` is
-positive or if `start` <= `end` and `step` is negative.
+positive or if `start <= end` and `step` is negative.
 
 ```yaml,json-e
 template:
--- a/docs/src/built-ins.md
+++ b/docs/src/built-ins.md
@@ -136,3 +136,27 @@ template: {$eval: 'len([1, 2, 3])'}
 context: {}
 result: 3
 ```
+
+## Range
+
+The `range()` built-in generates an array based on the following inputs:
+* `start` - An integer specifying the lower bound of the range (inclusive).
+  This can be negative.
+* `end` - An integer specifying the upper bound of the range (exclusive). This
+  can be negative.
+* `step` - Optional. An integer specifying a step to apply to each value within
+  the range. If not specified, defaults to `1`. Can be negative.
+
+For a positive step, the contents of a range r are determined by the formula
+r[i] = start + step*i where i >= 0 and r[i] < end.
+
+For a negative step, the contents of the range are still determined by the
+formula r[i] = start + step*i, where i >= 0 and r[i] > end.
+
+```
+template:
+  $map: {$eval: 'range(1, 5)'}
+  each(x): {$eval: 'x'}
+context:  {}
+result:   [1, 2, 3, 4]
+```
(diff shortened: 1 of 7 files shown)


djmitche (commented):
I think this is reasonable!

djmitche:
You'll probably want some error tests for passing non-integers or other invalid values to `range`.

gregsdennis:
@djmitche would the spec/docs also be updated with this PR?

djmitche:
Yep, I assume that's included in

> Update docs

:)

gregsdennis:
Ah, I see.  I'd have started with the spec before implementing it.

djmitche:
I just landed #536 to fix the go failures, btw

kwhitehouse:
Thanks @gregsdennis @djmitche !

@iadejumo and I have updated the PR to include implementation for all four languages. We've added more tests per your recommendation, including for negative ranges and non-integer parameters. We've also added a proposal for the `.md` documentation.

It's worth noting that adding support for `step` functionality and ranges of _decreasing_ value  added a decent amount of complexity in the Rust implementation. If you'd prefer we leave those out we can pull 'em.

Let us know what you think! We're happy to add more tests or adjust implementation if you folks have feedback.

specification.yml
```diff
@@ -1102,6 +1102,42 @@ template: {$map: {a: 1, b: 2, c: 3}, 'each(y)': {'${y.key}x': {$eval: 'y.val + 1
...
+context:  {}
+template:
+  $map: {$eval: 'range(10, 12)'}
+  each(x, i): 'id_${x}_${i}'
+result: ['id_10_0', 'id_11_1']
+---
+title:    simple range with step creates list of strings
+context:  {}
+template:
+  $map: {$eval: 'range(10, 20, 5)'}
```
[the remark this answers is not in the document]
kwhitehouse: Good point! I made a note in https://github.com/json-e/json-e/pull/534#issuecomment-2372488834 that the implementation for supporting `step` also was somewhat complex in Rust. If we want to leave it out for now because of that and the fact that it might be slightly unclear to users what these this third parameter will do, that's fine with us!

docs/src/built-ins.md
```diff
@@ -136,3 +136,27 @@ template: {$eval: 'len([1, 2, 3])'}
...
+
+## Range
+
+The `range()` built-in generates an array based on the following inputs:
+* `start` - An integer specifying the lower bound of the range (inclusive).
+  This can be negative.
+* `end` - An integer specifying the upper bound of the range (exclusive). This
+  can be negative.
+* `step` - Optional. An integer specifying a step to apply to each value within
+  the range. If not specified, defaults to `1`. Can be negative.
```
djmitche, line 148: But cannot be zero!

specification.yml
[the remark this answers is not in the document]
djmitche: It's not too complicated. I think we can leave it in. And, it's probably an unusual feature, so maybe the concern about readability isn't so much of an issue.

specification.yml
```diff
@@ -1102,6 +1102,110 @@ template: {$map: {a: 1, b: 2, c: 3}, 'each(y)': {'${y.key}x': {$eval: 'y.val + 1
...
+result:   0
+---
+title:    range with step creates list of numbers
+context:  {}
+template:
+  $map: {$eval: 'range(1, 10, 2)'}
+  each(x): {$eval: 'x'}
+result:   [1, 3, 5, 7, 9]
+---
+title:    range with negative step creates list of numbers
```
djmitche, line 1127: Please add a test for a negative step that doesn't end precisely on the end value, e.g., range(20, 0, -7).

rs/src/builtins.rs
```diff
@@ -188,6 +189,42 @@ fn strip_builtin(_context: &Context, args: &[Value]) -> Result<Value> {
...
+        // If step is not provided by the user, it defaults to 1.
+        None => 1,
+        Some(val) => match val {
+            Value::Number(n) if n.fract() == 0.0 => n.round() as i64,
+            _ => return Err(interpreter_error!("invalid arguments to builtin: range")),
+        }
+    };
+
+    if step > 0 {
+        let step = step as usize;
```
djmitche, line 216: `as` can be a little dangerous! The other uses here are infallible, but this could produce incorrect results on a 32-bit architecture. Of course, that architecture will fail when it tries to allocate more than 4GB of memory, but better a failure than incorrect results. You could use 
```
let step: usize = step.try_into()?;
```
to catch that case (maybe with `.map_err(..)`)

djmitche (commented):
Looks great!  A couple of minor notes here, but I'd expect this can be merged when those are addressed.

docs/src/built-ins.md
```diff
@@ -136,3 +136,27 @@ template: {$eval: 'len([1, 2, 3])'}
 context: {}
 result: 3
 ```
+
+## Range
+
+The `range()` built-in generates an array based on the following inputs:
+* `start` - An integer specifying the lower bound of the range (inclusive).
+  This can be negative.
```
gregsdennis, line 144: What does a negative value do?  (This should be explained.)

docs/src/built-ins.md
```diff
@@ -136,3 +136,27 @@ template: {$eval: 'len([1, 2, 3])'}
...
+* `end` - An integer specifying the upper bound of the range (exclusive). This
+  can be negative.
```
gregsdennis, line 146: Also list behavior here.

docs/src/built-ins.md
```diff
@@ -136,3 +136,27 @@ template: {$eval: 'len([1, 2, 3])'}
...
+* `step` - Optional. An integer specifying a step to apply to each value within
+  the range. If not specified, defaults to `1`. Can be negative.
+
+For a positive step, the contents of a range r are determined by the formula
+r[i] = start + step*i where i >= 0 and r[i] < end.
+
+For a negative step, the contents of the range are still determined by the
```
gregsdennis, line 153: Is it necessary to separate the positive and negative steps if they both use the same formula? 

Also, I found the inclusion of the pseudocode in the JSON Path spec to be quite helpful in implementing this logic (for JSON Path).  Maybe consider including that here, too.

specification.yml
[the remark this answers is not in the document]
gregsdennis: The typical construction for an array slice is 

```
[start]:[end][:[step]]
```

All values are optional. Defaults are

- start - 0
- end - the number of elements in the array
- step - 1

If step is included, the second colon is also required. 

This makes things like `3:`, `:10`, `:`, `::`, and `::2` all valid slices.

We don't have this complexity because we're our API a function, and the only optional parameter is step.  That simplifies things quite a bit.

I think it's pretty common to see them in this order.  As long as we maintain the common practice and document it, I think we'll be fine.

gregsdennis:
I think we're pretty covered with the current tests, but if you want more inspiration, check out [these tests](https://github.com/jsonpath-standard/jsonpath-compliance-test-suite/blob/main/tests%2Fslice_selector.json) from the JSON Path suite.

docs/src/built-ins.md
kwhitehouse: ✅  Added a note, good catch!

specification.yml
kwhitehouse: ✅  Great idea, done!

rs/src/builtins.rs
kwhitehouse: ✅  Updated to use `try_into()?`

I can update to make use of `map_err` if that's preferred - let me know. My Rust is rusty 😂 but it seems like the `map_err` would allow us to improve error messaging... is that correct?

docs/src/built-ins.md
kwhitehouse: ✅  Added content, please take a peek and let me know if you'd like other edits!

docs/src/built-ins.md
kwhitehouse: ✅  Added content, please take a peek and let me know if you'd like other edits!

docs/src/built-ins.md
kwhitehouse: ❔ I wasn't sure what you meant by the pseudocode in the JSON path spec! Happy to add more documentation here though - could you link to what you're referencing so I can use it as inspiration?

❔ You're right that both use the same formula! I kept separate because the `where` conditions are different. Since I updated this section, can you take another peek and let me know if you'd like further edits? Happy to make any changes to improve clarity / conciseness!

gregsdennis: Pseudocode is here: https://datatracker.ietf.org/doc/html/rfc9535#name-normative-semantics

kwhitehouse: Awesome! I added a block of pseudocode inspired by that link - thanks for the reference!

gregsdennis (approved):
I'm happy with the docs.  I'll leave the tooling implementations to @djmitche.

noahlaux (approved):

djmitche (approved):
This came together quickly and with a great result - thanks!

djmitche:
Releases take a little time, so I'll put it on my task list and get a new version out soon.

djmitche:
Oh, and I'll wait for #531 too.

djmitche:
v4.8.0 is released - thanks!

PR #542, opened 2024-10-19
Test and fix handling of false-y $default in $switch

### LINKED ISSUE #541: Empty strings in $default arm of switch operator are treated as null
This can be illustrated with a simple test case:

Example 1:
```
$switch:
  true: ''
  $default: ''
```

Example 2:
```
$switch:
  false: ''
  $default: ''
```

Here example 1 evaluates to `''`, and example 2 evaluates to `null`. I would expect them both to evaluate to an empty string.

This bug is made a bit worse by the fact that `json-e` omits the result of a `switch` operator from the parent container if it evaluates to `null`. Here's the (simplified) real world example where this was causing problems:

```
command:
  $let:
    extraArgs:
      $switch:
         'condition': '--extra'
         $default: ''
  in: >
    command --arg ${extraArgs}
```
Setting `condition` to false in the above example, results in:

> \# InterpreterError at template.command: unknown context value extraArgs

djmitche:
Tests revealed subtle failures here in both Go and JS. These bugs are unlikely to be relied on, so can be fixed at a bug-fix level.

Fixes #541.

### DIFF
--- a/internal/jsone.go
+++ b/internal/jsone.go
@@ -920,16 +920,17 @@ var operators = map[string]operator{
 			}
 		}
 
-		if len(result) == 0 && match["$default"] != nil {
-			value := match["$default"]
-			r, err := render(value, context)
-			if err != nil {
-				return nil, TemplateError{
-					Message:  err.Error(),
-					Template: template,
+		if len(result) == 0 {
+			if value, ok := match["$default"]; ok {
+				r, err := render(value, context)
+				if err != nil {
+					return nil, TemplateError{
+						Message:  err.Error(),
+						Template: template,
+					}
 				}
+				result = append(result, r)
 			}
-			result = append(result, r)
 		}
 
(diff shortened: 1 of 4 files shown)


djmitche:
@ahal, PTAL?

gregsdennis:
@djmitche does this affect any tests, or is it just implementation?

ahal:
Thanks for the fix, lgtm! (I didn't test it myself, but looks correct to me)

djmitche:
> @djmitche does this affect any tests, or is it just implementation?

It contains changes to specification.yml, if that's what you mean..

matt-boris (approved):
👍🏻

ahal (approved):

gregsdennis:
I've pulled this into my local, and it looks like JsonE.Net passes everything. 👍

PR #549, opened 2025-05-21
Fix golang lint warnings

### LINKED ISSUE #548: Go staticcheck failing
```
internal/interpreter/parser/parser.go:116:3: QF1003: could use tagged switch on token.Kind (staticcheck)
		if token.Kind == "[" {
		^
internal/jsone.go:291:17: ST1023: should omit type string from declaration; it will be inferred from the right-hand side (staticcheck)
		var seperator string = ""
		              ^
2 issues:
* staticcheck: 2
```

(also, it's spelled "separator"!)

djmitche:
Fixes #548.


djmitche:
(We don't specify a minimum go version, but this works with go1.20 at least)

petemoore (approved):
That looks nicer indeed, thanks Dustin!

PR #555, opened 2025-08-28
Rust: avoid panicking on non finite float

SheepTester:
This function seems to panic when it receives a non finite value, which is hard to recover from. Since its caller already returns a result, I made the function return a result too.

# Checklist

 I'll do these checklist items soon

### DIFF
--- a/rs/src/value.rs
+++ b/rs/src/value.rs
@@ -151,7 +151,7 @@ impl Value {
 /// Utility function to turn an f64 into a serde Number, using the simplest form.
 /// This conservatively assumes values outside (-u32::MAX..u32::MAX) are best
 /// represented as floats
-fn f64_to_serde_number(value: f64) -> Number {
+fn f64_to_serde_number(value: f64) -> Result<Number> {
     if value.fract() == 0.0 {
         if value < 0.0 && value > -(u32::MAX as f64) {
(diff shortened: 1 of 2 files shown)


rs/src/value.rs
```diff
@@ -151,16 +151,17 @@ impl Value {
...
+            return Ok((value as i64).into());
         } else if value >= 0.0 && value < u32::MAX as f64 {
-            return (value as u64).into();
+            return Ok((value as u64).into());
         }
     }
     // the failure conditions here are NaN and Infinity, which we do not see
-    Number::from_f64(value).unwrap()
+    Number::from_f64(value)
+        .ok_or_else(|| template_error!("{value} is not finite and cannot be represented in JSON"))
```
djmitche, line 164: I think this error can also happen for NaN? Perhaps just omit "is not finite", as the resulting error will then be either "Inf cannot be.." or "NaN cannot be.." which accurately describe the situation.

djmitche (commented):
This looks great - thank you! Just a tiny tweak to the error message here.

rs/src/value.rs
SheepTester: I agree, removed!

djmitche (approved):

PR #565, opened 2026-01-24
Update theme to be compatible with newer mdbook

djmitche:
The theme/index.hbs file is always a bit weird with mdbook: each version of mdbook has a "built-in" version, and if you want to customize it you copy that and make your changes. But then when you upgrade mdbook, it doesn't work.

As far as I could tell the only change was to add the logo, and I've re-applied that change here.

Fixes #564


petemoore (approved):
This looks good, many thanks Dustin!

PR #566, opened 2026-01-24
Update rust dependencies

djmitche:
This is the result of running `cargo update` to update packages within the ranges in `Cargo.toml`.


petemoore (approved):
Thanks Dustin!

PR #569, opened 2026-02-20
python: fix datetime.utcnow() deprecation warning

jcristau:
The `datetime.utcnow` method is deprecated since python 3.12.  `datetime.timezone.utc` was added in 3.2.


petemoore (approved):
Looks great, thanks!

PR #574, opened 2026-03-03
Add CLAUDE.md and expand CONTRIBUTING.md with project details

petemoore:
## Summary

- Move project overview, build/test commands, architecture details, and contribution conventions from `CLAUDE.md` into `CONTRIBUTING.md`, as suggested in review
- Preserve the original conversational style of `CONTRIBUTING.md` and weave in the new information alongside existing sections rather than replacing them
- Replace the Taskcluster-specific intro (bugsahoy, IRC, etc.) with a simpler welcome — json-e is no longer part of Taskcluster
- Add a minimal `CLAUDE.md` that points to `CONTRIBUTING.md`
- Gitignore `.claude/` (Claude Code local session data)

## Test plan

- [ ] Verify `CONTRIBUTING.md` reads well as a single coherent document
- [ ] Verify `CLAUDE.md` correctly points to `CONTRIBUTING.md`
- [ ] No functional code changes — documentation only

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,6 @@ env
 rs/target
 rs/**/*.rs.bk
 book
+
+# Claude Code local session data
+.claude/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+See CONTRIBUTING.md for project overview, build commands, architecture, and contribution conventions.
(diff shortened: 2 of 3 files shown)


djmitche (commented):
IMHO, documentation should be for humans and LLMs should be allowed to read it. So, let's move this information to CONTRIBUTING.md instead.

Then, refer to that from here in a way that claude can follow. We can then add similar `PLAUSIBLE-NONSENSE.md` files for other LLM vendors without duplicating content.

djmitche (approved):
Awesome: written for people, and the matrix multiplications can figure it out.

PR #585, opened 2026-07-10
fix: enforce spec error messages in the Go implementation

### LINKED ISSUE #393: [golang] enforce that error messages match
Currently the value of the `error: ..` property in `specification.yml` is ignored in the Go implementation, so the tests only assert that an error occurs, not that the message matches.

Let's fix that and make those error messages match.

mvanhorn:
In the Go implementation, `internal/jsone_test.go` loaded `specification.yml` and, for cases with an `error:` property, only asserted that *some* error occurred (`require.Error`), ignoring the expected message string. The Python harness already asserts the message exactly, so the spec strings are the canonical cross-language error messages.

This makes the Go spec test enforce the message: when a case's `error` is a string, `testCase.Test` now uses `require.EqualError(t, err, expected)` (keeping plain `require.Error` for `error: true`). It then brings the Go error messages into line with the spec: `TemplateError.Error()` renders the `TemplateError: <message>` form, the individual `TemplateError` and builtin messages (`BuiltinError: invalid arguments to builtin: <name>`) are aligned in `internal/jsone.go` and `internal/interpreter/interpreter.go`, interpreter messages in `internal/interpreter/execute.go` match the `InterpreterError: ...` forms, and `SyntaxError.Error()` in `internal/interpreter/parser/errors.go` matches the `SyntaxError: ...` format. All changes are string-level and fully specified by `specification.yml`.

# Checklist

* [x] Added a short description of the change to `newsfragments/393.bugfix`

Fixes #393

### DIFF
--- a/internal/interpreter/execute.go
+++ b/internal/interpreter/execute.go
@@ -18,6 +18,10 @@ var tokenizer = *p.NewTokenizer(`\s+`, strings.Split(
 	"null":       `null\b`,
 })
 
+var expectedInfixTokens = strings.Split(
+	`!= && ( * ** + - . / < <= == > >= [ in ||`, " ",
+)
+
 func Parse(source string, context interface{}) (interface{}, error) {
 	var parser p.Parser
 	var newInterpreter NewInterpreter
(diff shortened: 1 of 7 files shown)


djmitche (changes requested):
Apologies for the delay reviewing this - I was on holiday. This looks great! This has been a drawback to the Go implementation for quite some time so it's great to see it fixed.

Just a few requested changes below, and we can get this merged!

djmitche (approved):
Looks good -- thanks!!

PR #589, opened 2026-08-17
javascript impl: refactor to use dicts rather than objects

petemoore:
Treat contexts and template objects as explicit key/value dictionaries

### DIFF
--- a/js/src/index.js
+++ b/js/src/index.js
@@ -1,5 +1,3 @@
-/* eslint-disable */
-
 const {Parser} = require('./parser');
 const Tokenizer = require("../src/tokenizer");
 const {Interpreter} = require('./interpreter');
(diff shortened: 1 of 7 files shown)


js/src/index.js
Eijebong: Missed one

Eijebong: Missed another

Eijebong: And another

Eijebong:
```suggestion
  if (result.length === 0 && hasOwn(conditions, '$default')) {
```

Eijebong:
```suggestion
let res = mergeObjects(l);
for (let p of Object.keys(r)) {
  setProp(res, p, hasOwn(l, p) ? merge(l[p], r[p]) : r[p]);
}
```

Eijebong: Urgh... Well I don't know if it's github showing this wrong or if the review tool sucks. Anyway, this is a replacement for the whole `let res =...` and the for loop.

Eijebong (changes requested):
For all 2 out of the 3 `Object.assign` I marked:

You probably want to use something like this so it doesn't end up with a null __proto__ which would break since those objects are all fed back into json-e.

```js
let mergeObjects = (...sources) => {
  let result = {};
  for (let source of sources) {
    for (let key of Object.keys(source)) {
      setProp(result, key, source[key]);
    }
  }
  return result;
};
```

For the third one, my suggestion assumes you went with that

djmitche (approved):

PR #590, opened 2026-08-18
Treat object literals as explicit dicts and not objects

Eijebong:
This is a followup to d2bd496f4c8ab814e08fc25aebbb777c9bd92140 which missed that object literals had the same problem


petemoore (approved):
Solid fix, thanks Bastien!

PR #591, opened 2026-08-18
Use trusted publishing for releases

petemoore:
This moves to Trusted Publishing for releases.


release.sh
```diff
@@ -1,159 +1,174 @@
...
-            errors+=("Git remote '$REMOTE' is not reachable (check SSH keys / network)")
+    # Upstream main must be green before we release it. Taskcluster
+    # reports check-runs (reporting: checks-v1 in .taskcluster.yml), not
+    # legacy commit statuses, so query check-runs and ignore this
+    # workflow's own check (it hasn't run yet at this point anyway).
+    if [ -n "${remote_head}" ] && command -v curl >/dev/null 2>&1 && command -v jq >/dev/null 2>&1; then
+        local checks non_success
+        checks=$(curl -sS -H 'Accept: application/vnd.github+json' \
+                 "https://api.github.com/repos/json-e/json-e/commits/${remote_head}/check-runs?per_page=100" \
+                 2>/dev/null || echo '{}')
```
Eijebong: If github 500s you get no checks, which means the check after passes. Don't think that's what you wanted here

release.sh
```diff
@@ -1,159 +1,174 @@
...
+    local name="$1" url="$2" code
+    code=$(curl -sS -o /dev/null -w '%{http_code}' \
+           -H 'User-Agent: json-e-release-script (https://github.com/json-e/json-e)' \
+           "${url}" || echo "000")
+    case "${code}" in
+        200) echo yes ;;
+        404) echo no ;;
+        *)   echo "ERROR: ${name} returned HTTP ${code} for ${url}" >&2
+             echo "Cannot determine whether ${version} is already published." >&2
+             exit 1 ;;
```
Eijebong: This doesn't do what you think it does when you call `registry_has_version` in a subshell like in `    if [ "$(registry_has_version npm "https://registry.npmjs.org/json-e/${version}")" != no ]; then`

Only the subshell exits with 1. So you get both error messages. not the end of the world but not super great for debuggability.

scripts/check-versions.sh
```diff
@@ -0,0 +1,80 @@
...
+    else
+        printf '  ok        %-16s %s\n' "${what}" "${got}"
+    fi
+}
+
+echo "Checking all version strings are ${version}:"
+
+check "rs/Cargo.toml"   "$(sed -n 's/^version = "\(.*\)"$/\1/p' rs/Cargo.toml | head -1)"
+check "js/package.json" "$(sed -n 's/^  "version": "\(.*\)",$/\1/p' js/package.json | head -1)"
+check "py/setup.py"     "$(sed -n 's/^version = "\(.*\)"$/\1/p' py/setup.py | head -1)"
```
Eijebong: This is missing go? It's checked in the release workflow but that's a bit late for that since the tag is already published by then

release.sh
```diff
@@ -1,159 +1,174 @@
...
-version="${1}"
+#
+# Cut a json-e release.
+#
+# This script does NOT publish anything. It prepares and pushes the release
+# commit and annotated tag; pushing the tag triggers
+# .github/workflows/release.yml, which publishes to crates.io, npm and PyPI
+# via Trusted Publishing (OIDC) and then deploys the docs.
+#
+# See RELEASING.md.
```
Eijebong: release.sh: See RELEASING.md
REALEASING.md: `Run release.sh`

*spiderman_pointing.jpg*

.github/workflows/release.yml
```diff
@@ -0,0 +1,396 @@
...
+        if: steps.pre.outputs.skip == 'false'
+        with:
+          node-version: '24'
+          cache: yarn
+          cache-dependency-path: js/yarn.lock
+      - name: Ensure npm >= 11.5.1 (Trusted Publishing support)
+        if: steps.pre.outputs.skip == 'false'
+        run: |
+          set -euo pipefail
+          npm install -g npm@latest
```
Eijebong: Might want to pin that

release.sh
djmitche: GitHub would never 500

.github/workflows/release.yml
```diff
@@ -0,0 +1,382 @@
...
+
+env:
+  # Keep in sync with the mdbook version used for `MdBook tests` in
+  # .taskcluster.yml.
+  MDBOOK_VERSION: v0.5.2
+
+jobs:
+
+  # --------------------------------------------------------------------
+  # Gate 1: the tag is well-formed, cut from main, and every version
```
djmitche: The bots sure do like gates!

RELEASING.md
```diff
@@ -1,132 +1,50 @@
...
+**Never delete and re-create a `v*` tag.** `proxy.golang.org` caches
+`module@version` immutably within minutes of the first fetch; retagging gives
+every Go user who already fetched it a checksum-mismatch error with no way to
+recover. If a release goes wrong, bump the patch version and release again.
+
+## Integration audit
+
+Trusted Publishing and the `release` environment are already configured.
+This records what's live, for reference if it ever needs auditing or redoing
+(e.g. after a rename); it's not a to-do list.
```
djmitche: Weird LLMism here...

RELEASING.md
```diff
@@ -1,132 +1,50 @@
...
+
+All three registries bind trust to owner `json-e`, repo `json-e`, workflow
+filename `release.yml`, environment `release`. Renaming any of those breaks
+publishing to all three at once.
+
+| Registry | Configured via |
+|---|---|
+| [PyPI](https://pypi.org/manage/project/json-e/settings/publishing/) | Trusted publisher |
+| [npm](https://www.npmjs.com/package/json-e/access) | Trusted publisher |
+| [crates.io](https://crates.io/crates/json-e/settings) | Trusted Publishing |
```
djmitche: This probably isn't worth a table, since the second column is always the same

RELEASING.md
```diff
@@ -1,132 +1,50 @@
...
+the pushed tag *is* the release, resolved directly by `proxy.golang.org`.
+
+**Never delete and re-create a `v*` tag.** `proxy.golang.org` caches
+`module@version` immutably within minutes of the first fetch; retagging gives
+every Go user who already fetched it a checksum-mismatch error with no way to
+recover. If a release goes wrong, bump the patch version and release again.
+
+## Integration audit
+
+Trusted Publishing and the `release` environment are already configured.
```
djmitche: Could you provide a link to what this means?

djmitche (approved):

PR #592, opened 2026-08-20
Validate release packaging in CI and fix python packaging

petemoore:
Release 4.8.3 of python library failed due to missing `long_description_content_type`.

This fixes the issue, but also switches the python changelog from reStructuredText to (Github Flavored) Markdown to align with what we use for the rest of the docs (and .md seems like a nicer standard than .rst).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,9 +99,10 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
+      # Keep in sync with the version in .taskcluster.yml
       - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
-          python-version: '3.12'
+          python-version: '3.11'
       - name: Build sdist and wheel
         run: |
           set -euo pipefail
@@ -151,7 +152,11 @@ jobs:
           persist-credentials: false
       - name: Stage specification.yml into the crate
         run: ./scripts/stage-crate-spec.sh
-      - run: cargo --version && rustc --version
+      # ubuntu-latest's preinstalled Rust can lag behind whatever the
+      # runner image last baked in; update to track stable the same way
+      # .taskcluster.yml's rust:latest tasks do, rather than relying on
+      # an implicit, undated snapshot.
+      - run: rustup update stable && rustup default stable && cargo --version && rustc --version
       - name: cargo publish --dry-run
         working-directory: rs
         run: cargo publish --dry-run --allow-dirty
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -151,7 +153,11 @@ jobs:
           persist-credentials: false
       - name: Stage specification.yml into the crate
         run: ./scripts/stage-crate-spec.sh
-      - run: cargo --version && rustc --version
+      # ubuntu-latest's preinstalled Rust can lag behind whatever the
+      # runner image last baked in; update to track stable the same way
+      # .taskcluster.yml's rust:latest tasks do, rather than relying on
+      # an implicit, undated snapshot.
+      - run: rustup update stable && rustup default stable && cargo --version && rustc --version
```
Eijebong: "Fix python packaging", touches rust. Why?

.github/workflows/release.yml
```diff
@@ -99,9 +99,11 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
+      # Match the newest version .taskcluster.yml actually tests (3.8-3.11),
+      # rather than a newer one that's never been exercised by any CI here.
```
Eijebong: Most of this makes no sense outside of this diff. In the commit message not into some overly verbose comment

.taskcluster.yml
```diff
@@ -40,7 +40,8 @@ tasks:
                 git checkout ${repo.ref} &&
                 cd js &&
                 yarn &&
-                yarn test
+                yarn test &&
+                ./package-test.sh
```
Eijebong: Fix python packaging, touches js. Why?

.taskcluster.yml
```diff
@@ -94,7 +97,11 @@ tasks:
                 git config advice.detachedHead false &&
                 git checkout ${repo.ref} &&
                 cd rs/ &&
-                cargo test
+                cargo test &&
+                cd .. &&
+                ./scripts/stage-crate-spec.sh &&
+                cd rs &&
+                cargo publish --dry-run --allow-dirty
```
Eijebong: And again...

.github/workflows/release.yml
petemoore: Fair point, i've renamed PR title.

.github/workflows/release.yml
petemoore: We have fixed vesions burned into .taskcluster.yml and we want this file to be consistent. I think the comment is useful to justify why we are using e.g. python 3.11 and not python 3.12 here, so if someone were to update it, they would know to update .taskcluster.yml too.

petemoore: or maybe you are saying keep line 102 but drop line 103?

.taskcluster.yml
petemoore: See above - PR title renamed.

.taskcluster.yml
petemoore: And again same reply :-)

.github/workflows/release.yml
Eijebong: PR titles don't survive. Commits do. Unless you want to squash this which I would not do given the two commits are completely unrelated

.github/workflows/release.yml
Eijebong: `# Keep in sync with the version in .taskcluster.yml` or something like this.

Eijebong (approved):
Should've been 3 commits but whatever.

